PR #5, opened 2022-03-16
Fix obvious problems with Windows

stevapple:
Fix some broken or outdated Windows-related documentation.

### Motivation:

* Swift on Windows currently explicitly depends on Python 3.7. Reflect
this in our guide.
* Python version 3.7.8 doesn't exist in Windows Package Manager
repository. Point to 3.7.8150.0.
* Windows Package Manager can only install Swift 5.5.3 by the time, which
is obviously outdated.  Recommend Swift.org installers for now.

### Modifications:

* Fix Python dependency on Windows to 3.7
* Fix obvious typo in `getting-started/_repl.md`
* Fix incorrect `winget` package version
* Add Swift.org installer guide

### Result:

More precise guideline for installing and using Swift on Windows.

### DIFF
--- a/download/_windows.md
+++ b/download/_windows.md
@@ -38,9 +38,20 @@ start /w vs_community.exe --passive --wait --norestart --nocache ^
 del /q vs_community.exe
 ~~~
 
-1. Download the [latest package release](/download).
+0. Install Swift:
 
-2. Run the package installer, which will install a Swift toolchain into `%SystemDrive%\Library\Developer\Toolchains`.
+Swift can be installed through the official installer directly, or using the Windows Package Manager as well.  Notice that Windows Package Manager release may be behind the official release.
+
+* Using the official installer:
+  1. Download the [latest package release](/download).
+  1. Run the package installer.
+
+* Using the Windows Package Manager:
+  ~~~ cmd
+  winget install Swift.Toolchain
+  ~~~
+
+A Swift toolchain will be installed at `%SystemDrive%\Library\Developer\Toolchains\unknown-Asserts-development.xctoolchain`.  A compatible Swift SDK will be installed at `%SystemDrive%\Library\Developer\Platforms\Windows.platform\Developer\SDKs\Windows.sdk`.
 
 #### Traditional Installation
 
@@ -64,7 +75,7 @@ The following additional Visual Studio components are **recommended**:
 | Git for Windows | Microsoft.VisualStudio.Component.Git |
 | Python 3 64-bit (3.7.8) | Component.CPython.x64 |
 
-The following additional Visual Studio components are **suggested** for Swift 5.3:
+The following additional Visual Studio component is **suggested**:
 
 | Component | Visual Studio ID |
 |-----------|------------------|
--- a/download/_windows.md
+++ b/download/_windows.md
@@ -60,7 +60,7 @@ The following additional Visual Studio components are **recommended**:
 | Git for Windows | Microsoft.VisualStudio.Component.Git |
 | Python 3 64-bit (3.7.8) | Component.CPython.x64 |
 
-The following additional Visual Studio components are **suggested**:
+The following additional Visual Studio components are **suggested** for Swift 5.3:
 
 | Component | Visual Studio ID |
 |-----------|------------------|
--- a/download/_windows.md
+++ b/download/_windows.md
@@ -26,7 +26,7 @@ The platform dependencies cannot be installed through the Windows Package Manage
 
 ~~~ cmd
 winget install Git.Git
-winget install Python.Python.3 --version 3.7.8
+winget install Python.Python.3 --version 3.7.8150.0
 
 curl -sOL https://aka.ms/vs/16/release/vs_community.exe
 start /w vs_community.exe --passive --wait --norestart --nocache ^
--- a/download/_windows.md
+++ b/download/_windows.md
@@ -22,7 +22,9 @@ Windows has the following additional platform specific dependencies:
 
 The [Windows Package Manager](https://docs.microsoft.com/windows/package-manager/) can be found in the [App Store](https://www.microsoft.com/en-us/p/app-installer/9nblggh4nns1) or be [installed directly](ms-appinstaller:?source=https://aka.ms/getwinget).
 
-The platform dependencies cannot be installed through the Windows Package Manager as the install rules do not install the components necessary.
+0. Install required dependencies:
+ 
+The platform dependencies cannot be installed through the Windows Package Manager as the install rules do not install the components necessary.  They will be installed through Visual Studio installer.
 
 ~~~ cmd
 winget install Git.Git
@@ -34,9 +36,11 @@ start /w vs_community.exe --passive --wait --norestart --nocache ^
   --add Microsoft.VisualStudio.Component.Windows10SDK.19041 ^
   --add Microsoft.VisualStudio.Component.VC.Tools.x86.x64
 del /q vs_community.exe
-
-winget install Swift.Toolchain
 ~~~
+
+1. Download the [latest package release](/download).
+
+2. Run the package installer, which will install a Swift toolchain into `%SystemDrive%\Library\Developer\Toolchains`.
 
 #### Traditional Installation
 
--- a/download/_windows.md
+++ b/download/_windows.md
@@ -23,33 +23,33 @@ Windows has the following additional platform specific dependencies:
 The [Windows Package Manager](https://docs.microsoft.com/windows/package-manager/) can be found in the [App Store](https://www.microsoft.com/en-us/p/app-installer/9nblggh4nns1) or be [installed directly](ms-appinstaller:?source=https://aka.ms/getwinget).
 
 0. Install required dependencies:
- 
-The platform dependencies cannot be installed through the Windows Package Manager as the install rules do not install the components necessary.  They will be installed through Visual Studio installer.
 
-~~~ cmd
-winget install Git.Git
-winget install Python.Python.3 --version 3.7.8150.0
+   The platform dependencies cannot be installed through the Windows Package Manager as the install rules do not install the components necessary.  They will be installed through Visual Studio installer.
 
-curl -sOL https://aka.ms/vs/16/release/vs_community.exe
-start /w vs_community.exe --passive --wait --norestart --nocache ^
-  --installPath "%ProgramFiles(x86)%\Microsoft Visual Studio\2019\Community" ^
-  --add Microsoft.VisualStudio.Component.Windows10SDK.19041 ^
-  --add Microsoft.VisualStudio.Component.VC.Tools.x86.x64
-del /q vs_community.exe
-~~~
+   ~~~ cmd
+   winget install Git.Git
+   winget install Python.Python.3 --version 3.7.8150.0
+
+   curl -sOL https://aka.ms/vs/16/release/vs_community.exe
+   start /w vs_community.exe --passive --wait --norestart --nocache ^
+     --installPath "%ProgramFiles(x86)%\Microsoft Visual Studio\2019\Community" ^
+     --add Microsoft.VisualStudio.Component.Windows10SDK.19041 ^
+     --add Microsoft.VisualStudio.Component.VC.Tools.x86.x64
+   del /q vs_community.exe
+   ~~~
 
 0. Install Swift:
 
-Swift can be installed through the official installer directly, or using the Windows Package Manager as well.  Notice that Windows Package Manager release may be behind the official release.
+   Swift can be installed through the official installer directly, or using the Windows Package Manager as well.  Notice that Windows Package Manager release may be behind the official release.
 
-* Using the official installer:
-  1. Download the [latest package release](/download).
-  1. Run the package installer.
+   * Using the official installer:
+     1. Download the [latest package release](/download).
+     1. Run the package installer.
 
-* Using the Windows Package Manager:
-  ~~~ cmd
-  winget install Swift.Toolchain
-  ~~~
+   * Using the Windows Package Manager:
+     ~~~ cmd
+     winget install Swift.Toolchain
+     ~~~
 
 A Swift toolchain will be installed at `%SystemDrive%\Library\Developer\Toolchains\unknown-Asserts-development.xctoolchain`.  A compatible Swift SDK will be installed at `%SystemDrive%\Library\Developer\Platforms\Windows.platform\Developer\SDKs\Windows.sdk`.
 
--- a/getting-started/_installing.md
+++ b/getting-started/_installing.md
@@ -71,7 +71,9 @@ In order to develop applications, particularly with the Swift Package Manager, y
 
 The [Windows Package Manager](https://docs.microsoft.com/windows/package-manager/) can be found in the [App Store](https://www.microsoft.com/en-us/p/app-installer/9nblggh4nns1) or be [installed directly](ms-appinstaller:?source=https://aka.ms/getwinget).
 
-The platform dependencies cannot be installed through the Windows Package Manager as the install rules do not install the components necessary.
+0. Install required dependencies:
+ 
+The platform dependencies cannot be installed through the Windows Package Manager as the install rules do not install the components necessary.  They will be installed through Visual Studio installer.
 
 ~~~ cmd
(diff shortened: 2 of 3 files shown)


tomerd:
cc @compnerd

tomerd:
@swift-ci test

download/_windows.md
```diff
@@ -22,22 +22,26 @@ Windows has the following additional platform specific dependencies:
 
 The [Windows Package Manager](https://docs.microsoft.com/windows/package-manager/) can be found in the [App Store](https://www.microsoft.com/en-us/p/app-installer/9nblggh4nns1) or be [installed directly](ms-appinstaller:?source=https://aka.ms/getwinget).
 
-The platform dependencies cannot be installed through the Windows Package Manager as the install rules do not install the components necessary.
+0. Install required dependencies:
```
compnerd, line 25: Index by 1 please.

download/_windows.md
```diff
@@ -22,22 +22,26 @@ Windows has the following additional platform specific dependencies:
...
   --add Microsoft.VisualStudio.Component.Windows10SDK.19041 ^
   --add Microsoft.VisualStudio.Component.VC.Tools.x86.x64
 del /q vs_community.exe
-
-winget install Swift.Toolchain
 ~~~
 
+1. Download the [latest package release](/download).
+
+2. Run the package installer, which will install a Swift toolchain into `%SystemDrive%\Library\Developer\Toolchains`.
```
compnerd, line 43: These instructions are meant to explicitly use winget, please do not replace them.

download/_windows.md
```diff
@@ -60,7 +64,7 @@ The following additional Visual Studio components are **recommended**:
 | Git for Windows | Microsoft.VisualStudio.Component.Git |
 | Python 3 64-bit (3.7.8) | Component.CPython.x64 |
 
-The following additional Visual Studio components are **suggested**:
+The following additional Visual Studio components are **suggested** for Swift 5.3:
```
compnerd, line 67: Please keep this as is.  The suggestion terminology here is a standard thing (e.g. `apt get install` will also suggest items) - they are not required.

getting-started/_installing.md
```diff
@@ -71,22 +71,26 @@ In order to develop applications, particularly with the Swift Package Manager, y
 
 The [Windows Package Manager](https://docs.microsoft.com/windows/package-manager/) can be found in the [App Store](https://www.microsoft.com/en-us/p/app-installer/9nblggh4nns1) or be [installed directly](ms-appinstaller:?source=https://aka.ms/getwinget).
 
-The platform dependencies cannot be installed through the Windows Package Manager as the install rules do not install the components necessary.
+0. Install required dependencies:
```
compnerd, line 74: Index by 1 please.

getting-started/_installing.md
```diff
@@ -109,7 +113,7 @@ The following additional Visual Studio components are **recommended**:
 | Git for Windows | Microsoft.VisualStudio.Component.Git |
 | Python 3 64-bit (3.7.8) | Component.CPython.x64 |
 
-The following additional Visual Studio components are **suggested**:
+The following additional Visual Studio components are **suggested** for Swift 5.3:
```
compnerd, line 116: Same things as before apply.

getting-started/_repl.md
```diff
@@ -125,9 +125,9 @@ $R0: Int32 = 4
 
 ### On Windows
 
-The REPL depends on Python bindings.  You must ensure that Python is available
-in the path.  The following command adds Python to the PATH so that it can be
-used:
+The REPL depends on Python bindings.  You must ensure that Python 3.7 is available
+in the path.  The following command adds Python 3.7 from Visual Studio to `%PATH%`
+so that it can be used:
```
compnerd, line 130: Bah, I was leaving it unversioned intentionally, but you're right, there is no point.  The next snippet adds the version in explicitly :-(.

getting-started/_repl.md
```diff
@@ -138,6 +138,6 @@ extra parameters must be passed to the REPL.  This allows you to use multiple
 different SDKs with the same toolchain.
 
 ~~~ cmd
-set SWIFTFLAGS=-sdk %SDKROOT% -I %SDKROOT%/usr/lib/swift -L SDKROOT%/usr/lib/swift/windows
+set SWIFTFLAGS=-sdk %SDKROOT% -I %SDKROOT%/usr/lib/swift -L %SDKROOT%/usr/lib/swift/windows
```
compnerd, line 141: Nice catch!

compnerd (changes requested):

download/_windows.md
stevapple: Before we can make the latest Swift release available from winget, I don’t think it’s correct to teach others to do so.

getting-started/_installing.md
stevapple: This convention is picked up from the Linux part — can anyone responsible to explain this?

download/_windows.md
stevapple: The `winget` support is already broken. We cannot rely on this until we get Swift correctly versioned.

To be specific, we already have experimental RPMs for the latest release, but we’re not suggesting them on Swift.org. `winget` even doesn’t have that.

compnerd: I backed off from doing it as someone else picked up doing the version bump.  Otherwise, I normally do bump the releases once the official version is out.  I think that this is fine.

download/_windows.md
stevapple: Package suggestions can vary by version. Can you explain why we still "suggest" this in later releases?

getting-started/_installing.md
compnerd: I'm not sure what you mean by responsible here.

stevapple:
@tomerd I want to know if our guides are only intended for the latest release, or they should preserve maximum compatibility with older releases?

Swift on Windows has *much more legacy problems* due to missing optimization in early releases. For every release up to Swift 5.7, some of the restrictions or additional steps will get removed.

download/_windows.md
stevapple: `winget` maintainers appear to **reject** Swift from their releases because it lacks proper versioning, and CI is always failing.

compnerd: I've never really had issues with getting updates merged.

getting-started/_installing.md
stevapple: Just want to know why they’re doing that — I’ve double-checked and found we have every index to be `0` for Linux and `1` for Apple. Neither of them respects natural increments.

download/_windows.md
compnerd: I think that it is still useful.  Most features come in CMake before SPM.  This is the case on main as well (CMake supports more functionality).

download/_windows.md
stevapple: @compnerd Because `DiaplayVersion` validation was not a thing yet when you’re doing this.

Because we’re not able to add new releases any more, @svanimpe ended up covering Swift.Toolchain 5.5 with 5.5.3. But we cannot add Swift 5.6 at this time.

stevapple: Go and fix versioning first:(

If you already have that in mind, please share with me or @svanimpe. We’re pleased to take on the investigation.

download/_windows.md
stevapple: Okay, it makes some sense.

But I hope we can get rid of this soon when we have full support in bootstrapping SwiftPM and use it to rebuild all other stuffs, like what we’re doing on other platforms.

tomerd:
> @tomerd I want to know if our guides are only intended for the latest release, or they should preserve maximum compatibility with older releases?
>
> Swift on Windows has much more legacy problems due to missing fine-tuning in early releases. For every release up to Swift 5.7, some of the restrictions or additional steps will get removed.

imo the instructions should reflect the latest released version, or 5.6 at this point of time.

iirc we also keep track of instruction for older releases in a special section. @shahmishal ?

download/_windows.md
compnerd: I already have TODO notes for handling that.  The problem is how to figure out the version number in the wixproj file.  If you have an idea on how to resolve that, I'd love help/ideas.

download/_windows.md
compnerd: I don't think that I would rather keep the suggestion for Windows, at least for a while.

stevapple: So go on removing it, right?

getting-started/_installing.md
compnerd: Oh, I see what you mean.  If it's for symmetry with the Linux docs, okay, that is fair.

stevapple: Keeps with Linux doc style

download/_windows.md
stevapple: Glad to work on that. But before it happens, let’s fall back to installer.

We can always go back to winget when the support is done.

shahmishal:
> > @tomerd I want to know if our guides are only intended for the latest release, or they should preserve maximum compatibility with older releases?
> > Swift on Windows has much more legacy problems due to missing fine-tuning in early releases. For every release up to Swift 5.7, some of the restrictions or additional steps will get removed.
> 
> imo the instructions should reflect the latest released version, or 5.6 at this point of time.
> 
> iirc we also keep track of instruction for older releases in a special section. @shahmishal ?

@stevapple I would like to understand more about the difference between current (5.6) and previous release guidelines for Windows. We would like to keep the instruction for latest release.

stevapple:
@shahmishal AFAIK:
- Swift 5.3 vs 5.4: SwiftPM available; CMake no longer suggested.
- Swift 5.4 vs 5.5: installer capably of installing support files; copying support file no longer required.
- Swift 5.5 vs 5.6: I didn't notice internal changes yet. Swift 5.6 cannot be installed from `winget` as for now.
- Swift 5.6 vs 5.7: Move from VS2019 & Python3.7 to VS2022 & Python 3.9; `%SWIFTFLAGS%` no longer required.

compnerd:
> @shahmishal AFAIK:
> 
> * Swift 5.3 vs 5.4: SwiftPM available; CMake no longer suggested.


> * Swift 5.4 vs 5.5: installer capably of installing support files; copying support file no longer required.
> * Swift 5.5 vs 5.6: I didn't notice internal changes yet. Swift 5.6 cannot be installed from `winget` as for now.
> * Swift 5.6 vs 5.7: Move from VS2019 & Python3.7 to VS2022 & Python 3.9; `%SWIFTFLAGS%` no longer required.

stevapple:
> `%SWIFTFLAGS%` never was "required" - it was a way to make it easier for users to pass flags.

Right, but isn’t that Swift 5.7 with the new driver requires no additional flags at all?

> VS2022 has no meaning for the users.

UR right, I have reworded that.

stevapple:
Since winget will never be available for 5.6.0, I’m still on suggesting installer for now. We can change it back as soon as any compatible version is released.

compnerd:
> > `%SWIFTFLAGS%` never was "required" - it was a way to make it easier for users to pass flags.
> 
> Right, but isn’t that Swift 5.7 with the new driver requires no additional flags at all?
>

Correct, main does not require `-sdk %SDKROOT%`, but that is not the same thing.

> > VS2022 has no meaning for the users.
> 
> UR right, I have reworded that.

download/_windows.md
compnerd: The updates take time anyway, 5.6 was just released.  I think we should leave winget as is.

download/_windows.md
compnerd: Sorry, I meant, I would rather keep the suggestion.

download/_windows.md
stevapple: I don’t think we can get this into any released toolchain. Please always guide the user to use latest release, I don’t want to pollute this rule.

tomerd:
@stevapple @compnerd what are the next steps with this PR?

stevapple:
@tomerd IMO we agreed with each other except for the installation step. From my side, since Swift 5.6.0 will **never** gain winget support, we should guide users to use Swift.org installer until any winget-compatible version has released.

tomerd:
This kind of guide should reflect the current state of thing, i.e whichever option is best (safest, most reliable) to install 5.6. Once a new version of Swift is out, or some other material changes has occurred to this end, we should change the instructions accordingly. The two of you are the domain experts, so its up to you to decide what is the right way for 5.6. @ me on this PR when it is ready and I will get it merged.

stevapple:
> I think that we should keep the `winget` instructions; I would be okay with a note to say that it lags behind the installers on the website and referring them to that.

This cannot convince me since the guide seems only intended for the latest stable release. Why don’t we just restore `winget` instructions when the latest version is available again?

compnerd:
Because winget is the recommended way - it is effectively the package management system on Windows.  We should cater to that even if it lags behind.

tomerd:
in effort to get this over the finish line, how about the following:

=> regarding "which tool we use to install the toolchain": We mention both options and recommend the installer, e.g.

```
* Using the native installer: 
  * Download the [latest package release](/download).
  * Run the package installer, which will install a Swift toolchain into `%SystemDrive%\Library\Developer\Toolchains`.
* Using `winget` (may be behind the official release): `winget install Swift.Toolchain`
```

=> regarding CMake, I dont see a strong reason to drop it as a "suggested" tool. We could consider changing the "suggested" language to "optional" to make it softer.

wdyt?

stevapple:
> => regarding "which tool we use to install the toolchain": We mention both options and recommend the installer

Seems good.  I force-pushed a new version taking such path.

> => regarding CMake, I dont see a strong reason to drop it as a "suggested" tool. We could consider changing the "suggested" language to "optional" to make it softer.

I’m not against keeping it though. Will continue to mark it as suggested.

tomerd:
lgtm. @compnerd?

getting-started/_repl.md
compnerd, line 130: Weird line break here, I assume that this will get rendered appropriately.

compnerd (approved):

tomerd:
@swift-ci test

tomerd (approved):

download/_windows.md
```diff
@@ -22,22 +22,37 @@ Windows has the following additional platform specific dependencies:
...
 start /w vs_community.exe --passive --wait --norestart --nocache ^
   --installPath "%ProgramFiles(x86)%\Microsoft Visual Studio\2019\Community" ^
   --add Microsoft.VisualStudio.Component.Windows10SDK.19041 ^
   --add Microsoft.VisualStudio.Component.VC.Tools.x86.x64
 del /q vs_community.exe
-
-winget install Swift.Toolchain
 ~~~
 
+0. Install Swift:
```
tomerd, line 41: @stevapple what is `0.` here?

getting-started/_installing.md
```diff
@@ -71,22 +71,37 @@ In order to develop applications, particularly with the Swift Package Manager, y
...
 start /w vs_community.exe --passive --wait --norestart --nocache ^
   --installPath "%ProgramFiles(x86)%\Microsoft Visual Studio\2019\Community" ^
   --add Microsoft.VisualStudio.Component.Windows10SDK.19041 ^
   --add Microsoft.VisualStudio.Component.VC.Tools.x86.x64
 del /q vs_community.exe
-
-winget install Swift.Toolchain
 ~~~
 
+0. Install Swift:
```
tomerd, line 90: @stevapple what is `0.` here?

compnerd: It auto enumerates the index for the list.

tomerd:
Screen Shot 2022-03-23 at 5 55 07 PM

is this numbering expected? seems odd?

compnerd:
I suspect that there is something wrong with the indentation/wrapping that caused it to consider it to separate lists.

tomerd:
the README has instructions on how to run this locally, maybe worth double checking this looks right

stevapple:
@tomerd @compnerd Indentation problems are addressed.

tkremenek:
@swift-ci test

tomerd:
enumeration looks good now. thanks @stevapple @compnerd. next time please attach a screenshot of your changes - it can help avoid the extra iteration

tomerd:
@swift-ci test

PR #39, opened 2022-05-10
Fix Python version for Windows install

adam-fowler:
Change Python Install version for Windows Install dependencies

### Motivation:

Python install version is out of date for the current version of Swift

### Discussion:

@compnerd I haven't added the change here but I was wondering should we tell the user to open a new terminal after installing the dependencies and before installing swift. I can't be sure but I think that might be required to setup the paths correctly for the swift install.

### DIFF
--- a/download/_windows.md
+++ b/download/_windows.md
@@ -8,7 +8,7 @@ Swift has the following general dependencies:
 
 - Python[^1] (used by the debugger - lldb)
 
-[^1]: The windows binaries are built against Python 3.7.8
+[^1]: The windows binaries are built against Python 3.10.2
 
 Windows has the following additional platform specific dependencies:
 
@@ -38,7 +38,7 @@ The [Windows Package Manager](https://docs.microsoft.com/windows/package-manager
    del /q vs_community.exe
    ~~~
    
-1. Install Swift:
+0. Install Swift:
 
    Swift can be installed through the official installer directly, or using the Windows Package Manager as well.  Notice that Windows Package Manager release may be behind the official release.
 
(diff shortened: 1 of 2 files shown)


adam-fowler:
@compnerd Should we also include the `pip install six` call

download/_windows.md
```diff
@@ -37,14 +37,14 @@ The [Windows Package Manager](https://docs.microsoft.com/windows/package-manager
      --add Microsoft.VisualStudio.Component.VC.Tools.x86.x64
    del /q vs_community.exe
    ~~~
-
-0. Install Swift:
+
```
compnerd, line 40: Bleeding white space

download/_windows.md
```diff
@@ -37,14 +37,14 @@ The [Windows Package Manager](https://docs.microsoft.com/windows/package-manager
      --add Microsoft.VisualStudio.Component.VC.Tools.x86.x64
    del /q vs_community.exe
    ~~~
-
-0. Install Swift:
+   
+1. Install Swift:
```
compnerd, line 41: The 0. Is nicer - it auto indexes (unless the other instance got changed?)

compnerd (commented):
There is a separate copy in getting started IIRC, should that not be updated in parallel?

compnerd:
I’m not sure about `six`; I think that it is only used at build time, not runtime?  Was there something that used it at runtime?

As to the new terminal, Swift sets up the environment variables.  So it is only after that you need the new terminal to refresh the environment variables.  Are you concerned about `PYTHONHOME` or `PYTHONPATH`?  They aren’t referenced by the installer.

compnerd (approved):
Thanks!

adam-fowler:
> I’m not sure about `six`; I think that it is only used at build time, not runtime? Was there something that used it at runtime?

It is using it when I was getting the path to the lldb lib

shahmishal:
@swift-ci test

shahmishal (approved):

PR #92, opened 2022-06-21
Fix issues with property wrappers post

krstnfx:
### Motivation:

Fixes grammar issue and title display bug 

### Modifications:

Colons in Jekyll titles need to be escaped, but the previous workaround was causing a display issue in Safari. Escaped the title with quotes so that it would display properly.

### Result:
Screen Shot 2022-06-21 at 3 49 40 PM


twostraws (approved):

krstnfx:
@swift-ci test

PR #101, opened 2022-07-15
Fix VSCode Post Video

0xTim:
The VSCode Blog Post has a video pulled in via the markdown image tab. Only Safari supports loading videos via the image tab, so this converts it to a GIF so it works in all browsers


0xTim:
@swift-ci test

alexandersandberg (approved):
Looks good to me. 👍 The quality is slightly lower with the GIF, but I think it's fine.

An alternative is using a `video` tag with `muted` and `autoplay` properties.

```markdown
<video muted autoplay>
  <source src="{{ site.baseurl }}/assets/images/vscode-extension-blog/symbol-completion.mp4">
</video>
```

PR #118, opened 2022-08-19
Adding `<meta name="description" />` for docs

sanjaiyan-dev:
Adding [`<meta name="description" />`](https://web.dev/meta-description)

_Sorry if I made any mistake :(_

### DIFF
--- a/_layouts/base.html
+++ b/_layouts/base.html
@@ -4,6 +4,7 @@
 <head>
   <meta charset="utf-8" />
   <title>{{ site.title }} - {{ page.title | escape }}</title>
+  <meta name="description" content="{{ page.excerpt | strip_html | escape }}" />
   <meta name="author" content="Apple Inc." />
   <meta name="viewport" content="width=device-width initial-scale=1" />
   <link rel="license" href="/LICENSE.txt" />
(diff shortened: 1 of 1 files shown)


alexandersandberg:
What would you like to accomplish with these changes? I know it's best practice to add a `description` tag for SEO purposes, but it's not a _requirement_ for good SEO.

While I'm sure Swift.org's SEO can be improved, looking at the [Google search results for the website](https://www.google.com/search?q=swift.org+site:swift.org), the descriptions generated by Google (because of the omitted `description` tag) seem fine to me.

_layouts/base.html
```diff
@@ -4,6 +4,7 @@
 <head>
   <meta charset="utf-8" />
   <title>{{ site.title }} - {{ page.title | escape }}</title>
+  <meta name="description" content="{{ page.excerpt | strip_html | escape }}" />
```
alexandersandberg, line 7: `page.excerpt` is not available for all pages, which would result in an empty `description` tag.

The Twitter meta tags fall back to using `site.description`, but for the `description` tag, it would be better to omit it instead and let the search engine generate its own description based on the contents of the page.

sanjaiyan-dev: > `page.excerpt` is not available for all pages, which would result in an empty `description` tag.
> 
> The Twitter meta tags fall back to using `site.description`, but for the `description` tag, it would be better to omit it instead and let the search engine generate its own description based on the contents of the page.


Ok 🙌🏽

sanjaiyan-dev:
https://github.com/apple/swift-org-website/pull/118#issuecomment-1223713535

PR #127, opened 2022-09-02
Update landing page content

kaishin:
### Modifications:

- First pass of updating the content of the landing page.

### DIFF
--- a/_layouts/default.html
+++ b/_layouts/default.html
@@ -3,7 +3,9 @@ layout: base
 ---
 
 <main role="main">
-  {{ content }}
+  <article class="page">
+    {{ content }}
+  </article>
 </main>
 
 {% include footer.html %}
--- a/index.md
+++ b/index.md
@@ -11,7 +11,7 @@ Swift is a general-purpose programming language built using a modern approach to
 <div class="links links-list-nostyle" markdown="1">
   - [Get started with Swift](/getting-started)
   - [Read the documentation](/documentation)
-  - [Latest release: 5.6.3](/download/#swift-563)
+  - [Latest release: 5.7](/download/#swift-57)
 </div>
 
 ## Why Swift?
--- a/index.md
+++ b/index.md
@@ -6,9 +6,82 @@ atom: true
 
 # Welcome to Swift.org
 
-Welcome to the Swift community. Together we are working to build a programming language to empower everyone to turn their ideas into apps on any platform.
+Swift is a general-purpose programming language built using a modern approach to safety, performance, and software design patterns.
 
-Announced in 2014, the Swift programming language has quickly become one of the fastest growing languages in history. Swift makes it easy to write software that is incredibly fast and safe by design. Our goals for Swift are ambitious: we want to make programming simple things easy, and difficult things possible.
+<div class="links links-list-nostyle" markdown="1">
+  - [Get started with Swift](/getting-started)
+  - [Read the documentation](/documentation)
+  - [Latest release: 5.6.2](/download/#swift-562)
+</div>
 
-For students, learning Swift has been a great introduction to modern programming concepts and best practices. And because it is open, their Swift skills will be able to be applied to an even broader range of platforms, from mobile devices to the desktop to the cloud.
- 
+## Why Swift?
+
+### Safe
+
+The most obvious way to write code should also behave in a safe manner. Undefined behavior is the enemy of safety, and developer mistakes should be caught before software is in production. Opting for safety sometimes means Swift will feel strict, but we believe that clarity saves time in the long run.
+
+### Fast
+
+Swift is intended as a replacement for C-based languages (C, C++, and Objective-C). As such, Swift must be comparable to those languages in performance for most tasks. Performance must also be predictable and consistent, not just fast in short bursts that require clean-up later. There are lots of languages with novel features — being fast is rare.
+
+### Expressive
+
+Swift benefits from decades of advancement in computer science to offer syntax that is a joy to use, with modern features developers expect. But Swift is never done. We will monitor language advancements and embrace what works, continually evolving to make Swift even better.
+
+## Use cases
+
+### Apple platforms
+
+Swift is a powerful and intuitive programming language on macOS, iOS, watchOS, tvOS and beyond. Swift code is safe by design, yet also produces software that is optimized to run lightning-fast.
+
+<div class="links links-list-nostyle" markdown="1">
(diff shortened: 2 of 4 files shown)


kaishin:
cc @alexandersandberg

index.md
```diff
@@ -6,9 +6,82 @@ atom: true
...
 # Welcome to Swift.org
 
-Welcome to the Swift community. Together we are working to build a programming language to empower everyone to turn their ideas into apps on any platform.
+Swift is a general-purpose programming language built using a modern approach to safety, performance, and software design patterns.
 
-Announced in 2014, the Swift programming language has quickly become one of the fastest growing languages in history. Swift makes it easy to write software that is incredibly fast and safe by design. Our goals for Swift are ambitious: we want to make programming simple things easy, and difficult things possible.
+<div class="links links-list-nostyle" markdown="1">
+  - [Get started with Swift](/getting-started)
+  - [Read the documentation](/documentation)
+  - [Latest release: 5.6.2](/download/#swift-562)
```
tomerd, line 14: => can we generate this from the metadata available cc @shahmishal 
=> eventually will this link to the "install" page which will point to the latest version?

shahmishal: I will be working on adding the metadata to make this easier. When it's ready I will post the PR here.

alexandersandberg: > => eventually will this link to the "install" page which will point to the latest version?

If it's a "latest release" link I think it makes sense to link to /download, where the latest version would be highlighted/separated at the too.

A link to the future /install page should be readily available as well though, imo.

0xTim (approved):
Looks like a great start!

index.md
kaishin: I was indeed trying to figure out if there is a way to infer this from metadata.
Should I wait for @shahmishal's PR to reduce churn?

_layouts/default.html
```diff
@@ -3,7 +3,9 @@
 ---
 
 <main role="main">
-  {{ content }}
+  <article class="page">
+    {{ content }}
+  </article>
```
alexandersandberg, line 8: Is `<article>` semantically correct here? Is it better to change the CSS rules to not require an `<article>` tag? (I assume this is for the list arrows.)

index.md
```diff
@@ -6,9 +6,82 @@ atom: true
...
+Swift benefits from decades of advancement in computer science to offer syntax that is a joy to use, with modern features developers expect. But Swift is never done. We will monitor language advancements and embrace what works, continually evolving to make Swift even better.
+
+## Use cases
+
+### Apple platforms
+
+Swift is a powerful and intuitive programming language on macOS, iOS, watchOS, tvOS and beyond. Swift code is safe by design, yet also produces software that is optimized to run lightning-fast.
+
+<div class="links links-list-nostyle" markdown="1">
+  - [Learn more about Swift on Apple platforms](#)
```
alexandersandberg, line 38: Maybe we should add some TODOs to the code to not forget about (and possibly merge) things that need to be fixed? (update the link, in this case)

index.md
```diff
@@ -6,9 +6,82 @@ atom: true
...
+  - [Learn more](/contributing/#triaging-bugs)
+</div>
+
+## What's new
+
+Stay up-to-date with the latest in the Swift community.
+
+<div class="links links-list-nostyle" markdown="1">
+  - [Visit the Swift.org blog](/blog/)
+  - [Learn about the Swift forums](/community/#forums)
```
alexandersandberg, line 85: Might be better to link to the forums directly?

index.md
dempseyatgithub: Since the current patch version number of Swift often doesn't match up with the version of Swift included with the latest Xcode. I wonder if this will be a source of confusion.

i.e. Someone thinks their version of Swift is 'out of date' because they have 5.6.1 as part of Xcode 13.4 — instead of 5.6.3 which was released on Sept 2.

_layouts/default.html
kaishin: Regarding `article`, it [can be nested](https://developer.mozilla.org/en-US/docs/Web/HTML/Element/article) in any [flow element](https://developer.mozilla.org/en-US/docs/Web/Guide/HTML/Content_categories), which includes `main`.

The CSS-only approach was the first one I tried. Top level styles did the job, but it's hard to review unintended side effects. Making the styles scoped to the landing page felt off as well. Happy to reconsider if we see any drawbacks to treating the landing page like any other page with an `article` tag wrapping the content.

alexandersandberg: I believe `<article>` should only be used for e.g. article _items_ in a list and not the whole article itself. For this case, where the content is the main part of the page, I believe `<main>` only would be the correct way to do it—even though nesting `<article>` might still be allowed.

I might be wrong, but that's what I can remember.

daveverwer: I think `article` would not cause any real-world problems here, but `section` can also be a useful tag in this situation.

kaishin: > I believe `<article>` should only be used for e.g. article _items_ in a list and not the whole article itself. For this case, where the content is the main part of the page, I believe `<main>` only would be the correct way to do it—even though nesting `<article>` might still be allowed.
> 
> I might be wrong, but that's what I can remember.

Indeed, but it is not exclusive to these case. The following structure is fairly common and is semantically correct when used with single-article pages:

```html
<main>
  <article>
  </article>
</main>
```

Putting the semantic aspect aside, the reason I used `article` is because we're already using it for this purpose in other layouts, the goal being to minimize CSS changes in the branch until the content starts shaping up a bit more.

alexandersandberg: Ok, great, then we can resolve this conversation! 🙂 Thanks for the explanations.

alexandersandberg (approved):

_layouts/default.html
kaishin: Great discussion to have! Resolving for now.

index.md
```diff
@@ -11,7 +11,7 @@ Swift is a general-purpose programming language built using a modern approach to
 <div class="links links-list-nostyle" markdown="1">
   - [Get started with Swift](/getting-started)
   - [Read the documentation](/documentation)
-  - [Latest release: 5.6.2](/download/#swift-562)
+  - [Latest release: 5.6.3](/download/#swift-563)
```
alexandersandberg, line 14:
```suggestion
  - [Latest release: 5.7](/download/#swift-57)
```

5.7 now even! 🙂

PR #136, opened 2022-09-12
Fix typo in the Download page about Swift 5.7

shahmishal:


shahmishal:
@swift-ci test

alexandersandberg (approved):
Is this note needed even? The table already specifies Xcode 14. Maybe we can remove it.

shahmishal:
Normally use it for Linux/Windows only release

alexandersandberg:
@shahmishal Ok, then maybe we should fix the windows footnote reference, as there are currently two 1s.

shahmishal:
Good catch!

PR #150, opened 2022-09-22
Replace "iff" in API guidelines

amartini51:
Developers unfamiliar with the use of "iff" as an abbreviation for "if and only if" are confused and file bugs, both against this page and against reference for the Swift standard library, because it looks like a typo.

This PR re-creates <https://github.com/apple/swift-internals/pull/20>, which was closed during the Git repository move.  For an example of real documentation with and without "iff", see <https://github.com/apple/swift/pull/39977>.

Fixes rdar://56670458

### DIFF
--- a/documentation/api-design-guidelines/index.md
+++ b/documentation/api-design-guidelines/index.md
@@ -638,13 +638,16 @@ is printed.
 
   ~~~ swift
   extension Shape {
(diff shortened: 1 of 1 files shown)


alexandersandberg (approved):

alexandersandberg:
@swift-ci please test

PR #32, opened 2022-04-25
Update the Bug tracker for Swift to GitHub Issues

shahmishal:
### Motivation:

Swift bugs are moving from bugs.swift.org to GitHub Issues. 

### Modifications:

* Update all bugs.swift.org urls to GitHub Issues. 
* Remove out of date StarterBugs list and provide direct link to the label.

PR #151, opened 2022-09-25
Update name of bugtracker

Kyle-Ye:
After #32 , we update the bug track link to Github Issues.

And since swift-org-website is open sourced too. Maybe "the Swift.org bug tracking system" should be redirected to this repo instead of apple/swift repo.

### DIFF
--- a/community/index.md
+++ b/community/index.md
@@ -12,7 +12,7 @@ The Swift language is developed in the open, and all technical or administrative
 
 * Directory of forum categories and email instructions are in the [forum section](#forums)
 * Source code for all Swift projects can be found on GitHub at [github.com/apple][github]
-* The Swift.org bug tracking system is maintained at [github.com/apple/swift/issues][bugtracker]
+* The Swift language bug tracking system is maintained at [github.com/apple/swift/issues][bugtracker]
 
 All communication within project spaces should adhere to Swift project's [Code of Conduct](/code-of-conduct).
 


Kyle-Ye:
cc @shahmishal

alexandersandberg (commented):
This Communication section is about the Swift _language_ project and not the Swift.org website project. See the first sentence:

> The Swift language is developed in the open, ...

So the link is correct. However, I think we can improve the text here.

> "The **Swift.org bug tracking system**..."

I can see why this might be understood as "where we track Swift.org bugs", which is not the case.

Maybe removing the ".org" is better?

> The Swift bug tracking system is maintained at...

community/index.md
```diff
@@ -12,7 +12,7 @@ The Swift language is developed in the open, and all technical or administrative
 
 * Directory of forum categories and email instructions are in the [forum section](#forums)
 * Source code for all Swift projects can be found on GitHub at [github.com/apple][github]
-* The Swift.org bug tracking system is maintained at [github.com/apple/swift/issues][bugtracker]
+* The Swift.org bug tracking system is maintained at [github.com/apple/swift-org-website/issues][bugtracker]
```
alexandersandberg, line 15: Note that this change would only update the link _text_, and not the link itself. To update the `bugtracker` link, you also have to make a change on [L135](https://github.com/apple/swift-org-website/blob/df0040d67a37fc0c75fbaa4dfe2bb1089e664e5b/community/index.md?plain=1#L135).

Kyle-Ye: Nice catch

Kyle-Ye:
> This Communication section is about the Swift _language_ project and not the Swift.org website project. See the first sentence:
> 
> > The Swift language is developed in the open, ...
> 
> So the link is correct. However, I think we can improve the text here.
> 
> > "The **Swift.org bug tracking system**..."
> 
> I can see why this might be understood as "where we track Swift.org bugs", which is not the case.
> 
> Maybe removing the ".org" is better?
> 
> > The Swift bug tracking system is maintained at...

+1 for removing the ".org" since this section seems to be mainly focus on Swift itself rather than swift-org-website.

But changing the link is just a bug-fix while changing the content is not. Maybe this should be double checked by more code owners?

alexandersandberg:
I believe the link is already correct by linking to github.com/apple/swift/issues (the Swift language project's bug tracking system). It should not link to the Swift.org website project's repo.

Kyle-Ye:
> I believe the link is already correct by linking to github.com/apple/swift/issues (the Swift language project's bug tracking system). It should not link to the Swift.org website project's repo.

Got it. I think we should better remove the misleading ".org" to make them consistent

alexandersandberg (approved):
Reading over this again, would it be even more clear to explicitly mention the Swift language project?

> The Swift _language_ bug tracking system is maintained at...

The reason is that the source code for _all_ projects are mentioned in the list item above, changing the context a bit.

Or maybe we can place the list item mentioning all projects at the end of the list (swap 2nd and 3rd)?

Now I'm probably nitpicking, so I'm approving this either way. 😄

Kyle-Ye:
@swift-ci please test

alexandersandberg:
@swift-ci please test

Kyle-Ye:
> Reading over this again, would it be even more clear to explicitly mention the Swift language project?
> 
> > The Swift _language_ bug tracking system is maintained at...
> 
> The reason is that the source code for _all_ projects are mentioned in the list item above, changing the context a bit.
> 
> Or maybe we can place the list item mentioning all projects at the end of the list (swap 2nd and 3rd)?
> 
> Now I'm probably nitpicking, so I'm approving this either way. 😄

Thanks for the review feedback. Changed to "Swift Language".

PR #155, opened 2022-09-29
[sswg] Fix date format

97mik:


0xTim (approved):
As much as I dislike this - thanks!

The SWWG should probably discuss what standard we want to stick to because this date format is confusing to everyone outside the US

0xTim:
@swift-ci test

alexandersandberg:
+1 for a more international date format. 😁

Kyle-Ye:
> The international standard recommends writing the date as year, then month, then the day: YYYY-MM-DD.
> Source: https://en.wikipedia.org/wiki/Date_format_by_country

Should we use align them to be something like 2022-09-29?

0xTim:
> Should we align them all to be something like 2022-09-29?

Yeah we need to discuss this and then audit the site for all the other instances of the ~wrong~ American formatting

PR #162, opened 2022-10-12
SEO improvement

jcabak:
- Document have a meta description (Now we have 100/100 points in Lighthouse scoring in SEO :) )

### DIFF
--- a/_layouts/base.html
+++ b/_layouts/base.html
@@ -9,6 +9,7 @@
   {% endif %}
   <meta name="author" content="Apple Inc." />
   <meta name="viewport" content="width=device-width initial-scale=1" />
+  <meta name="description" content="{{ page.excerpt | strip_html | escape }}">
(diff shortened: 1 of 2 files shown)


daveverwer:
Thanks very much for your contribution, @jcabak, and apologies for the extremely long time you’ve been waiting for a response!

I’ve taken a look at this today, and the problem you have noticed here is part of a bigger problem with the website’s meta tags in general. 

This PR now correctly shows a page `excerpt` if one is defined. While this will help with blog posts on the site, no other pages have excerpts. However, this is a step in the right direction, and I think we should merge it. 👍 

I have logged the other issues in #203.

daveverwer:
@swift-ci please test

daveverwer (approved):

daveverwer:
Thanks so much for your contribution, @jcabak 👍

PR #166, opened 2022-10-25
Fixed an old mailing list link

daveverwer:
It looks like this link was missed during the switch from mailing lists to forums. The page still exists, but the anchor has gone so the visitor was left at the top of the community page.


daveverwer:
@swift-ci please test

0xTim (approved):

shahmishal (approved):

PR #184, opened 2022-12-20
Add data file with items for "Go Further" section

dempseyatgithub:
Add data file with items for _Go Further_ section of _Getting Started_ page.

This PR is against the `content-improvements` branch.

As discussed in previous workgroup meetings, here is the initial pass at data for items for the Go Further section.

Note that the thumbnail URLs are TBD but included as placeholder entries.

### DIFF
--- a/_data/go_further.yml
+++ b/_data/go_further.yml
@@ -17,7 +17,7 @@
   thumbnail-url: #TBD
   release-date: 2022-06-07
 - title: "Go Further: Swift Strings Under The Hood"
-  description: "Swift strings provide a fast, Unicode-compliant way to work with text. In most cases"
+  description: "Swift strings provide a fast, Unicode-compliant way to work with text. This article provides more detail on how strings are implemented in Swift using UTF-8 and the reasons behind that choice."
   content-type: article
   content-url: https://www.swift.org/blog/utf8-string/
   thumbnail-url: #TBD
--- a/_data/go_further.yml
+++ b/_data/go_further.yml
@@ -2,23 +2,23 @@
   description: "Swift has concurrency features built into the language making it easier to write concurrent code with the assistance of the compiler. This video introduces the async/await mechanism, a key part of Swift Concurrency."
   content-type: video
   content-url: https://developer.apple.com/videos/play/wwdc2021/10132/
-  thumbnail-url: TBD
+  thumbnail-url: #TBD
   release-date: 2021-06-08
 - title: "Go Further: Swift Generics"
   description: "In Swift, generics are a fundamental way to write abstract code. This video walks through the basics of Swift Generics and introducing generics into your code."
   content-type: video
   content-url: https://developer.apple.com/videos/play/wwdc2021/10132/
-  thumbnail-url: TBD
+  thumbnail-url: #TBD
   release-date: 2021-06-09
 - title: "Go Further: Regular Expressions"
   description: "Swift provides first-class regular expression support, commonly known as regex, for effective string processing. This video gives an overview of the power and flexibility of Swift Regex."
   content-type: video
   content-url: https://developer.apple.com/videos/play/wwdc2022/110357/
-  thumbnail-url: TBD
+  thumbnail-url: #TBD
   release-date: 2022-06-07
 - title: "Go Further: Swift Strings Under The Hood"
   description: "Swift strings provide a fast, Unicode-compliant way to work with text. In most cases"
   content-type: article
   content-url: https://www.swift.org/blog/utf8-string/
-  thumbnail-url: TBD
+  thumbnail-url: #TBD
   release-date: 2019-03-20
--- a/_data/go_further.yml
+++ b/_data/go_further.yml
@@ -1,0 +1,24 @@
+- title: "Go Further: Concurrency"
+  description: "Swift has concurrency features built into the language making it easier to write concurrent code with the assistance of the compiler. This video introduces the async/await mechanism, a key part of Swift Concurrency."
+  content-type: video
+  content-url: https://developer.apple.com/videos/play/wwdc2021/10132/
+  thumbnail-url: TBD
+  release-date: 2021-06-08
+- title: "Go Further: Swift Generics"
+  description: "In Swift, generics are a fundamental way to write abstract code. This video walks through the basics of Swift Generics and introducing generics into your code."
+  content-type: video
+  content-url: https://developer.apple.com/videos/play/wwdc2021/10132/
+  thumbnail-url: TBD
+  release-date: 2021-06-09
+- title: "Go Further: Regular Expressions"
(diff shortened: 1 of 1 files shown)


_data/go_further.yml
```diff
@@ -0,0 +1,24 @@
+- title: "Go Further: Concurrency"
+  description: "Swift has concurrency features built into the language making it easier to write concurrent code with the assistance of the compiler. This video introduces the async/await mechanism, a key part of Swift Concurrency."
+  content-type: video
+  content-url: https://developer.apple.com/videos/play/wwdc2021/10132/
+  thumbnail-url: TBD
```
tomerd, line 5: considering leaving empty so that the rendering engine can ignore it easily

_data/go_further.yml
```diff
@@ -0,0 +1,24 @@
...
+  release-date: 2021-06-09
+- title: "Go Further: Regular Expressions"
+  description: "Swift provides first-class regular expression support, commonly known as regex, for effective string processing. This video gives an overview of the power and flexibility of Swift Regex."
+  content-type: video
+  content-url: https://developer.apple.com/videos/play/wwdc2022/110357/
+  thumbnail-url: TBD
+  release-date: 2022-06-07
+- title: "Go Further: Swift Strings Under The Hood"
+  description: "Swift strings provide a fast, Unicode-compliant way to work with text. In most cases"
+  content-type: article
```
tomerd, line 21: consider "blog-post" to make more explicit

_data/go_further.yml
dempseyatgithub: I've made the TBD a comment so the value is null and easily ignored but still describes the intent to have a value.

_data/go_further.yml
dempseyatgithub: I had added the `content-type` field in case we wanted to display types of content differently. The example I had in mind was if we wanted to display videos inline.

I didn't have a use case in mind where we would display text content differently depending on whether it was an article or blog post, so 'article' seemed like a good general term.  Do you think we would be displaying articles vs blog posts differently?

I also thought about using 'text' but felt post or article conveyed what it was better could have images, diagrams, etc.

tomerd: I think there is a case to be made that the rendering of the item depends on the "content-type", in that sense a link to a blog post on the website could be different from an external article

tomerd (approved):

tomerd:
cc @TimTr to review (or pull the right person) if these are the right resources and topics

_data/go_further.yml
dempseyatgithub: I think there is probably a difference between "content-type" which was intended for the presentation medium (video, article, possibly audio such as a podcast episode) and something like "content-source" which could be whether an item is from an official Swift source (swift.org, Apple) or an external source.

One could imagine linking to an external blog post, which would then make a content type of 'blog-post' ambiguous. I was also trying to avoid hair-splitting over when something is considered an article vs a blog post.

One reason 'article' seemed a good choice is that it mirrors the semantic HTML tag 'article' which specifies articles, blog posts, and forum posts as some of its stated example uses.

Maybe it makes sense to add something like a "content-source" field? If so, would values of 'official' and 'external' suffice for now? Or maybe an "is-external" flag would suffice.

Since all of the entries are from official sources for the foreseeable future, maybe adding a field should be deferred until we put together the guidelines for linking to external resources?

dempseyatgithub:
> cc @TimTr to review (or pull the right person) if these are the right resources and topics

We had reviewed these items at a workgroup meeting that I don't think @TimTr was able to be at.

To give a little more context, the three videos for concurrency, generics, and regex are good introductions to intermediate Swift topics that represent recent advancements in Swift.

Concurrency and generics are also topics that developers will encounter when working with frameworks like SwiftUI or Vapor.

The blog post about Swift String and UTF-8 is included as an example of an article-based item. It is an excellent article, but may be too advanced for the _Go Further_ section.

Since the _Go Further_ section of the _Getting Started_ page will be dynamically generated from this data, the intent was to have a representative article entry for layout / design purposes, even if this particular item is commented out before the new _Getting Started_ page goes live.

_data/go_further.yml
kaishin: For the time being it seems like there isn't much to gain from adding such a flag. Perhaps in the future we can add a flag that explicitly marks community-contributed resources instead of the other way around. That approach should be clearer in theory compared to "external" since it's reasonable to argue that apple.com links are external to swift.org.

kaishin: Regarding the name, I also prefer `article` over `blog-post`. It would leave the door open to link  to gists, blog posts, forum posts, etc. without requiring us to define a type for each.

_data/go_further.yml
```diff
@@ -0,0 +1,24 @@
+- title: "Go Further: Concurrency"
```
kaishin, line 1: I'm wondering if including the "Go Further" in the title would constrain the way these can be displayed. Any drawbacks of dropping it from here and leaving it to the presentation layer?

_data/go_further.yml
```diff
@@ -0,0 +1,24 @@
...
+  thumbnail-url: #TBD
+  release-date: 2021-06-09
+- title: "Go Further: Regular Expressions"
+  description: "Swift provides first-class regular expression support, commonly known as regex, for effective string processing. This video gives an overview of the power and flexibility of Swift Regex."
+  content-type: video
+  content-url: https://developer.apple.com/videos/play/wwdc2022/110357/
+  thumbnail-url: #TBD
+  release-date: 2022-06-07
+- title: "Go Further: Swift Strings Under The Hood"
+  description: "Swift strings provide a fast, Unicode-compliant way to work with text. In most cases"
```
kaishin, line 20: Is the abrupt sentence ending intentional here?

kaishin (commented):
A couple of suggestions related to the metadata, otherwise looking good!

_data/go_further.yml
dempseyatgithub: I definitely see your point, but was thinking that we should avoid making things too general right now and wait until we have a second concrete use case for these entries, so we know what generalization makes sense.

For example just plain "Concurrency" or "Swift Generics" are very general titles on their own. So if these entries were to be used in another context, it would be important not to just use the 'title' as-is.

_data/go_further.yml
dempseyatgithub: Yes, this may have been a copy/paste victim moving the text from my draft to the data file. Will add a commit.

_data/go_further.yml
kaishin: Fair enough. 👍

_data/go_further.yml
dempseyatgithub: I think we will leave it as `article` for now, but of course can always revisit in the future if necessary.

PR #198, opened 2023-01-09
swift-evolution: update search text input

MaxDesiatov:
### Motivation:

Search field on the Swift Evolution page has default user agent style applied, which makes it too small and close to unreadable on Safari. In other browsers the field does not fit with the website style, as it has no border radius and also has inconsistent position.

#### Safari:

*Before:*

safari_before

*After:*

safari_after

#### Firefox

*Before:*

firefox_before

*After:*

firefox_after

#### Edge

*Before:*

edge_before

*After:*

edge_after

### Result:

Search field on the Swift Evolution page is readable and fits the overall website design in all browsers.

### DIFF
--- a/assets/stylesheets/pages/_swift-evolution.scss
+++ b/assets/stylesheets/pages/_swift-evolution.scss
@@ -5,6 +5,16 @@
         flex-direction: row;
         flex-wrap: wrap;
         justify-content: space-between;
+        height: 3.5rem;
+    }
+
+    #search-filter {
+        font-size: .8rem;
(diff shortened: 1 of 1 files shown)


MaxDesiatov:
@swift-ci test

tomerd (approved):
nice, thank you

kaishin (approved):
Thanks for addressing this. Much better outcome and can serve as the foundation for all future text fields we introduce.

shahmishal (approved):

PR #200, opened 2023-01-10
Update the steps to select toolchain in Xcode

shahmishal:
https://github.com/apple/swift-org-website/issues/199

Issue #199, opened 2023-01-09
Installation instructions are wrong at https://www.swift.org/download/


dabrahams:
Quoting from that page:

> Download the [latest package release](https://www.swift.org/download).

Well, presumably these instructions should work for any of the toolchains published, yeah?

> Run the package installer, which will install an Xcode toolchain into /Library/Developer/Toolchains/.
> An Xcode toolchain (.xctoolchain) includes a copy of the compiler, lldb, and other related tools needed to provide a cohesive development experience for working in a specific version of Swift.

It's not clear that it does; at least, having installed a 5.8 development snapshot, this command finds nothing:

```
 find  /Library/Developer/Toolchains/swift-5.8-DEVELOPMENT-SNAPSHOT-2022-12-29-a.xctoolchain -name swiftc
 ```

> Open Xcode’s Preferences, navigate to Components > Toolchains, and select the installed Swift toolchain.

There's no "Preferences" in XCode.  There's "Settings," but no "Components" in there.  You want "XCode > Toolchains"

> Selecting a Swift toolchain affects the Xcode IDE only. To use the Swift toolchain with command-line tools, use xcrun --toolchain swift and xcodebuild -toolchain swift,

nope:
```
$ xcrun --toolchain swift
xcrun: error: no utility name specified
Usage: xcrun [options] <tool name> ... arguments ...
```

also not this:

```
xcrun --toolchain /Library/Developer/Toolchains/swift-5.8-DEVELOPMENT-SNAPSHOT-2022-12-29-a.xctoolchain swift --version
```
(still chooses Swift 5.7)

Dredging my memory for how this is actually done, I open the Info.plist inside the  toolchain package, find a magic toolchain identifier, and try this:

```
$ xcrun --toolchain org.swift.58202212291a swift --version
xcrun: error: can't exec '/Library/Developer/Toolchains/swift-5.8-DEVELOPMENT-SNAPSHOT-2022-12-29-a.xctoolchain/usr/libexec/swift' (errno=Permission denied)
$ TOOLCHAINS=org.swift.58202212291a xcrun swift --version
xcrun: error: can't exec '/Library/Developer/Toolchains/swift-5.8-DEVELOPMENT-SNAPSHOT-2022-12-29-a.xctoolchain/usr/libexec/swift' (errno=Permission denied)
```

>  or add the Swift toolchain to your path as follows:
>
[...]

PR #201, opened 2023-01-11
download: Misc improvements to package installation steps on Apple platforms

AnthonyLatsis:
This is mostly an iteration on #200 and is motivated by #199.

* Don’t suggest downloading just the latest release package.
* Draw attention to the importance of meeting the system requirements for a particular package.
* Uppercase "LLDB" for consistency across the website.
* Don’t suggest navigating to `> Manage Toolchains...`; the toolchain can be selected directly from `Xcode > Toolchains`.
* We already define the installed toolchain as an "Xcode toolchain". Use "installed toolchain" rather than "Swift toolchain" to avoid cognitive overhead.
* Use "default toolchain" instead of "Xcode toolchain" where it refers to the toolchain that ships with Xcode, and not the format.
* Bring more clarity to how to use the installed toolchain with command line tools and provide a relevant example.


MaxDesiatov:
I think there's also some confusion on whether the Swift package can be installed and is fully functional without Xcode. I've seen reports that not everything works, but may be we should call this out in the document? That is, if it only partially works without Xcode, we should highlight that?

AnthonyLatsis:
It doesn’t sound unreasonable to adjust for a "no Xcode" path if Xcode is indeed not required to run the installer. Do you have a use case in mind?

>  I've seen reports that not everything works

You should be able to run the binaries in an installed toolchain directly over anything that doesn’t depend on an SDK. Is that what you meant by "partially"?

MaxDesiatov:
I was referring to this report specifically https://forums.swift.org/t/spm-support-missing-on-monterey-without-xcode-no-such-module-packagedescription/61001

0xTim:
There's also https://github.com/apple/swift-package-manager/issues/4396

AnthonyLatsis:
> maybe we should call this out in the document?

How does this look?

AnthonyLatsis:
ping

download/_apple-platforms.md
```diff
@@ -53,23 +59,40 @@ To submit to the App Store you must build your app using the version of Swift th
...
 
-1. Selecting a Swift toolchain affects the Xcode IDE only.  To use the Swift toolchain with command-line tools, use `xcrun --toolchain swift` and `xcodebuild -toolchain swift`, or add the Swift toolchain to your path as follows:
+  Alternatively, you may add the installed toolchain to your path as follows:
 
-~~~ shell
-$ export PATH=/Library/Developer/Toolchains/swift-latest.xctoolchain/usr/bin:"${PATH}"
-~~~
+  ~~~ shell
+  export PATH=/Library/Developer/Toolchains/swift-latest.xctoolchain/usr/bin:"${PATH}"
+  ~~~
```
MaxDesiatov, line 95: There's [one more way](https://samwize.com/2022/05/23/how-to-use-a-different-swift-toolchain-in-xcode/) to temporarily enable a specific toolchain with `export TOOLCHAINS`, which seems a bit less disruptive than `export PATH`. It's a bit more involved though, as it requires the end user to lookup `CFBundleIdentifier` of an already installed toolchain.

MaxDesiatov: Here's a one-liner that could make it easier for the user:
```
export TOOLCHAINS=$(plutil -extract CFBundleIdentifier raw /Library/Developer/Toolchains/swift-DEVELOPMENT-SNAPSHOT-2023-01-16-a.xctoolchain/Info.plist)
```

AnthonyLatsis: A couple of questions:
* Why less disruptive? (I’m actually curious)
* This seems to me like an exploitation of an implementation detail. Can we be sure this won’t stop working all of a sudden with a future release of, say, `xcrun`?
* I take it `export TOOLCHAINS` would select a specific toolchain only for invocations of `xcrun` and `xcodebuild`, unlike `export PATH`. Is there any reason other than convenience that the reader would want to use this instead of passing the bundle identifier as an argument to `--toolchain`?

MaxDesiatov: > Why less disruptive? (I’m actually curious)

Overloading `PATH` brings in everything into scope from that directory, while `TOOLCHAINS` only exposes Swift-related tools. I feel much more comfortable tweaking Swift-specific environment purposes when trying out the new toolchain than modifying something more fundamental like `PATH`.

>This seems to me like an exploitation of an implementation detail. Can we be sure this won’t stop working all of a sudden with a future release of, say, xcrun?

I think this is also applicable to Xcode in general. The ability to select toolchains via Xcode's Settings window has been removed recently, so all of these steps (including Xcode GUI guidance) would have to be re-evaluated for every major Xcode release anyway, I suppose.

>I take it `export TOOLCHAINS` would select a specific toolchain only for invocations of `xcrun` and `xcodebuild`, unlike export PATH.

That's not correct, it will make `/usr/bin/swift` (which is itself just a shim) redirect to correct executables, which means means users don't have to prepend `xcrun --toolchain` to their every Swift-related command invocation. At the same time they can preserve their `PATH` settings intact.

AnthonyLatsis: Yay, `export TOOLCHAINS` can also be set to a toolchain alias!

AnthonyLatsis:
@shahmishal ping

MaxDesiatov (approved):

shahmishal:
@swift-ci test

shahmishal (approved):
Thanks!

PR #206, opened 2023-01-17
[SE Dashboard] Update the search-bar height style

benrimmington:
Follow-up to apple/swift-org-website#198.

### Motivation:

The search-bar was recently changed to use a fixed height, which is a rendering issue when the filter-options are expanded.

### Modifications:

Use `min-height` instead of `height`.

### Result:

* Before (with `height`):
  Before

* After (with `min-height`):
  After

### DIFF
--- a/assets/stylesheets/pages/_swift-evolution.scss
+++ b/assets/stylesheets/pages/_swift-evolution.scss
@@ -5,7 +5,7 @@
         flex-direction: row;
         flex-wrap: wrap;
         justify-content: space-between;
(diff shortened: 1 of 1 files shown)


MaxDesiatov:
@swift-ci test

MaxDesiatov (approved):
Great catch, thanks!

alexandersandberg (approved):
Thanks!

shahmishal (approved):

PR #208, opened 2023-01-18
Getting started tweaks

daveverwer:
### Motivation:

Improving the Getting Started page from the perspective of someone completely new to Swift.

### Modifications:

* Removed the references to `swift build`, `swift test`, `swift run`. If someone had just installed Swift to the point where they can run `swift --version` as recommended on the lines above, these commands are going to fail while looking for `Package.swift`. The use case examples below introduce these commands in a more natural way.
* Tweaked how we introduce the SwiftPM abbreviation when we first introduce it (and also on the package manager page).
* Link to the package manager page when introducing SwiftPM.
* Added a link to TSPL. I know we want to get into some slightly less daunting tasks than "Read this whole book", but I feel it's important to have a link to it somewhere on this page. I tried to phrase it as optional.

### DIFF
--- a/_includes/getting-started/_installing.md
+++ b/_includes/getting-started/_installing.md
@@ -4,4 +4,4 @@ If you don't have Swift installed, [install it first](/install).
 
 To test that you have Swift installed, run `swift --version` from your shell or terminal app.
 
-Swift comes bundled with the Swift Package Manager (SwiftPM), which manages the distribution of your Swift code, and allows easy importing of other Swift packages into your apps.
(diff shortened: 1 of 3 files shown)


tomerd (approved):

alexandersandberg (approved):
This looks great!

daveverwer:
@swift-ci please test

I'm not sure what exactly this is going to test as it's not being merged into `main` yet and it'll all get a test when that merge finally happens, but I figure it's better safe than sorry!

PR #216, opened 2023-01-23
Fix navigation link in `navigation.yml`

MaxDesiatov:
The order of links https://www.swift.org/contributing page is currently wrong, as "Website and Blog Post Contributions" section on the page comes before "Swift Evolution". The link itself to `#blog-post-contributions` is not working, since the actual anchor id is `website-and-blog-post-contributions`.

### DIFF
--- a/_data/navigation.yml
+++ b/_data/navigation.yml
@@ -99,10 +99,10 @@
           name: reporting-bugs
         - title: Triaging Bugs
           name: triaging-bugs
+        - title: Blog Post Contributions
+          name: website-and-blog-post-contributions
         - title: Swift Evolution
           name: evolution-process
-        - title: Blog Post Contributions
-          name: blog-post-contributions
         - title: Contributing Code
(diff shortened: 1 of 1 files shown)


MaxDesiatov:
@swift-ci test

tomerd (approved):

tomerd:
@shahmishal is there a "broken link" test we can add to CI to catch these?

shahmishal:
We don't have any, it would be great to add it.

daveverwer:
Thanks for spotting and fixing this, Max! You're right it started life in a different position and with a slightly different name. 😬

> is there a "broken link" test we can add to CI to catch these?

We could absolutely do it, but as they're anchors rather than individual pages, we might need a bit of custom code written to check them, as Jekyll won't fail for a broken anchor.

shahmishal:
https://github.com/apple/swift-org-website/issues/217

PR #231, opened 2023-02-02
SSWG: Cassandra client is accepted

ktoso:
The proposal review of https://forums.swift.org/t/sswg-0021-swift-cassandra-client/62658 concluded and accepted the library

PR #232, opened 2023-02-03
Generate Projects list on Swift on Server page with data file

alexandersandberg:
### Motivation:

I [noticed recently](https://github.com/apple/swift-org-website/pull/231#issuecomment-1413283423) that the projects table on the Swift on Server page was using a static table instead of the existing (but outdated) projects data file.

Using a data file will make the process of maintaining this list easier.

### Modifications:

- Updated the `_data/swwg/projects.yml` file to match the static table on the Swift on Server page
- Replaced the static table with one generated from the data file

Note: The date format is now changed from MM/DD/YYYY to YYYY-MM-DD (05/16/2019 → 2019-05-16). As we've had some issues with the former before (mixing up days and months), I think this more internationally recognized date format is better.

### Result:

The same table as before but now generated from the data file. When SSWG now wants to update this list they should change the `_data/sswg/projects.yml` file instead.

As I was manually updating the data file based on the data in the static table, feel free to help me double-check that all the new data is correct. 🙂

cc: @ktoso @0xTim @tomerd 

CleanShot 2023-02-03 at 09 47 10

### DIFF
--- a/_data/sswg/projects.yml
+++ b/_data/sswg/projects.yml
@@ -1,77 +1,160 @@
 - name: SwiftNIO
+  description: Event-driven network application framework.
   maturity: Graduated
-  pitched: 2018-09-07
+  pitched: N/A
   accepted: 2018-09-07
   url: http://github.com/apple/swift-nio/
 
-- name: Logging API
+- name: SwiftLog
+  description: Logging API
   maturity: Graduated
   pitched: 2018-09-10
   accepted: 2019-02-07
   url: http://github.com/apple/swift-log/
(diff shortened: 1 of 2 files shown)


ktoso (approved):
LGTM, awesome! I didn't know we can use templating here after all -- great to have it generated instead, thanks 👍

alexandersandberg:
@swift-ci please test

tomerd (approved):
nice one, thanks!

kaishin:
🙌

Issue #261, opened 2023-03-21
Searching for "Feature Identifier" on swift.org/swift-evolution


MarcoEidinger:
## Is your feature request related to a problem? Please describe

Proposals may have a new, optional field that defines the feature identifier that allows for [Piecemeal adoption of upcoming language improvements](https://github.com/apple/swift-evolution/blob/cf8419c15503dc2384b469c29b64c20bc905ea6c/proposals/0362-piecemeal-future-features.md)

Unfortunately, it is not possible to search for proposals that have a feature identifier. It is also not possible to search for a proposal with a given feature identifier. 

## Describe the solution you'd like

- Being able to search and find proposals that have a "Feature Identifier" (a.k.a. upcoming feature flag)
- Being able to search and find a proposal for a given Feature Identifier

## Additional context

Excerpt from [Proposals define their own feature identifier ](https://github.com/apple/swift-evolution/blob/cf8419c15503dc2384b469c29b64c20bc905ea6c/proposals/0362-piecemeal-future-features.md#proposals-define-their-own-feature-identifier)

> [Swift proposal template](https://github.com/apple/swift-evolution/blob/main/proposal-templates/0000-swift-template.md) got amended with a new, optional field that defines the feature identifier:

> **Feature identifier**: UpperCamelCaseFeatureName
> Amend the following proposals, which are partially or wholly delayed until Swift 6, with the following feature identifiers:

> Several proposals were amended to mention the respective feature identifier

> * [SE-0274 "Concise magic file names"](https://github.com/apple/swift-evolution/blob/main/proposals/0274-magic-file.md) (`ConciseMagicFile`) delayed the semantic change to `#file` until Swift 6. Enabling this feature changes `#file` to mean `#fileID` rather than `#filePath`.
> * [SE-0286 "Forward-scan matching for trailing closures"](https://github.com/apple/swift-evolution/blob/main/proposals/0286-forward-scan-trailing-closures.md) (`ForwardTrailingClosures`) delays the removal of the "backward-scan matching" rule of trailing closures until Swift 6. Enabling this feature removes the backward-scan matching rule.
> * [SE-0335 "Introduce existential `any`"](https://github.com/apple/swift-evolution/blob/main/proposals/0335-existential-any.md) (`ExistentialAny`) delays the requirement to use `any` for all existentials until Swift 6. Enabling this feature requires `any` for existential types.
[...]

PR #268, opened 2023-03-28
Add support for feature identifier field on Swift Evolution page

dempseyatgithub:
- Read and display `upcomingFeatureFlag` field if present
- Enable search on that field
- Use a map between proposal id and upcoming feature flags temporarily until backing JSON file adds the `upcomingFeatureFlag` field.
- Part of Issue #261 

This PR adds support for an 'upcoming feature flag' field to be read, displayed, and searched on the Swift Evolution page.

Note that the `upcomingFeatureFlag` field is not present in the source `proposals.json` file at the moment. This PR uses a map from proposal id to upcoming feature flag in the JavaScript file to add the property to the appropriate proposals. This is intended to be temporary until the property is included in the JSON.

Note also that this Issue #261 also includes filtering for all proposals with an upcoming feature flag. This PR does not address that part of the issue.

### DIFF
--- a/assets/javascripts/swift-evolution.js
+++ b/assets/javascripts/swift-evolution.js
@@ -349,6 +349,7 @@ function renderProposals() {
       if (state === '.implemented') detailNodes.push(renderVersion(proposal.status.version))
       if (state === '.previewing') detailNodes.push(renderPreview())
       if (proposal.implementation) detailNodes.push(renderImplementation(proposal.implementation))
+      if (proposal.featureIdentifier) detailNodes.push(renderFeatureIdentifier(proposal.featureIdentifier))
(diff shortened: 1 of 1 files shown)


dempseyatgithub:
@swift-ci test

dempseyatgithub:
Since it probably doesn't make sense to merge this until proposals.json is returning the new field, I turned this into a draft. Also, it is likely the name of the field will be `upcomingFeatureFlag` instead of `featureIdentifier` to match the name being used in the existing proposals that have this field.

dempseyatgithub:
Updated the PR to the new name `upcomingFeatureFlag` and added a map from proposal id to flag to make this immediately useful without waiting for changes in the source json file.

Moving from draft to be able to review.

dempseyatgithub:
@swift-ci test

shahmishal (approved):

PR #274, opened 2023-03-31
Improve page metadata for sharing

### LINKED ISSUE #203: Meta tag, Open Graph, and Twitter sharing metadata problems
As part of looking over whether we should merge #162, I noticed we have some problems with how we are using our meta tags, including Open Graph and Twitter tags.

When people share URLs from Swift.org, they currently use the site description for the Open Graph and Twitter meta tags, meaning the Mentorship page previews using the site title and site description:

There are several things we could do better:

* Tweak the conditional meta tags around lines 36 and 49 of `base.html`, only falling back to the site title and description when absolutely nothing else is available.
* Attempt to get auto-excerpts working for pages that are not blog posts. I’m not quite sure how they are working for blog posts right now since we don’t use an `excerpt_separator`. This needs a little investigation.
* Every page should already have a sensible title, so even without automated excerpts, we should change it so we use the page title in Open Graph and Twitter meta tags.
* Stretch Goal: We should consider adding custom explicit excerpts for pages that get linked to a lot, like the workgroup pages, getting started, and community pages.

dempseyatgithub:
This PR addresses the issues in Issue #203:

- In _config.yml file:
  - Enable page excerpts
  - Set excerpt separator
- In base layout:
  - Calculate title, description, URL in one place for meta tags
  - Always use page values for meta tags when present, fallback to site values
  - Replace conditional tags with single tag that uses calculated value
  - Move robots meta tag out from the middle of Open Graph tags
  - Reduce excessive whitespace from generated newlines
- Add explicit excerpts to front matter of pages where automatic excerpt was not suitable / sufficient
- Small adjustments to page first paragraphs to generate better description meta tags
- Resolves #203

### DIFF
--- a/_config.yml
+++ b/_config.yml
@@ -5,6 +5,8 @@ timezone: America/Lower_Princes
 exclude: ["README.md", "config.ru", "Gemfile", "Gemfile.lock", "Procfile", "vendor"]
 safe: true
 future: true
+excerpt_separator: <!--end excerpt-->
+page_excerpts: true
 
 kramdown:
     syntax_highlighter: rouge
--- a/_layouts/base.html
+++ b/_layouts/base.html
@@ -1,5 +1,29 @@
 <!DOCTYPE html>
 <html lang="en">
+  
+{%- if page.title -%}
+  {%- capture meta_title -%}
+    {{ page.title | escape }}
+  {%- endcapture -%}
+{%- else -%}
+  {%- assign meta_title = site.title -%}
+{%- endif -%}
+
+{%- if page.excerpt -%}
+  {%- capture meta_description -%}
+    {{ page.excerpt | strip_html | strip | escape }}
+  {%- endcapture -%}
+{%- else %}
+  {%- assign meta_description = site.description -%}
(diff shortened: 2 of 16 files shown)


dempseyatgithub:
@swift-ci test

dempseyatgithub:
I just found an unexpected issue with this PR.

It causes all the posts on the main blog page to render full text instead of excerpts.

I won't have time today to track down the cause, but will turn this back into a full PR once I address that issue.

dempseyatgithub:
I have spent some time investigating the issue.

There seems to be a bug in Jekyll where turning on page excerpts causes each post on the blog page to display its full text and not its excerpt.

I tested with the latest version of Jekyll and still see this issue.

I also have found the issue to be nondeterministic. Sometimes I can enable page excerpts without the bug occurring, sometimes it seems adding a call into the page excerpt for the meta tags causes it to happen. However, backing out the changes does not always revert the behavior.

I haven't found a similar issue mentioned searching in Jekyll forums. My next step would be to try to reproduce the issue in a simple Jekyll project.

PR #276, opened 2023-03-31
Remove SE-0374 from Swift 5.8 post

dempseyatgithub:
- The proposal is now listed as being implemented in 5.9

There was some back and forth over this issue in the original post PR.

The proposal is now listed on the evolution page as implemented in 5.9, so it should be removed from the Swift 5.8 post.

Here is the link to the proposal: https://www.swift.org/swift-evolution/#?proposal=SE-0374


dempseyatgithub:
@swift-ci test

tkremenek (approved):

Issue #281, opened 2023-04-02
Previous Swift releases hard to find


SimplyDanny:
## Describe the bug

Previous Swift versions are indented which lets them appear like belonging to the RPM section instead of the general [Releases](https://www.swift.org/download/#releases).

## Expected behavior

These older Swift releases should be at the same level like the [Swift 5.8 release](https://www.swift.org/download/#swift-58).

## Actual behavior

They are nesting and it looks like they belong to the previous CentOS 7 section.

## Screenshots

![Screenshot 2023-04-02 at 15 27 25](https://user-images.githubusercontent.com/16365760/229356090-29bf4ff5-382e-417f-8d95-e417f50d13c1.png)

### Desktop (please complete the following information):

- OS: macOS 13.3
- Browser: Safari
- Version: 16.4

### Swift.org version/commit hash

1069e001ebccad8dfb742a98efda7855b1b1374d

alexandersandberg:
Hey @SimplyDanny, we've made some improvements on this in our content improvements initiative which is now ready for review: https://github.com/apple/swift-org-website/pull/354

We would appreciate your feedback there! 🙂

PR #282, opened 2023-04-02
Use headlines with correct level instead of emphasis

SimplyDanny:
* "RPM" now belongs to "Swift 5.8" instead of being on the same level. 
* Using headlines instead of emphasis for "Amazon Linux 2" and "CentOS 7" is generally advised.


daveverwer:
@swift-ci please test

daveverwer (approved):
I’ve tested this locally and the change is good, but I still think the previous releases section merges into this as you mention in #281.

Maybe we should have an `h3` either above or as part of the `_older-releases.md`

/cc @shahmishal

SimplyDanny:
Yes, I know. It is not the intention of this PR to fix #281. I also tried to fix it but without success.

0xTim:
Will merge this and we can fix #281 in another PR

PR #291, opened 2023-04-27
link to developer.apple.com from apple use case

tomerd:
motivation: handle Apple use cases

changes: link to developer.apple.com from apple use case link, remove TODO


tomerd:
cc @alexandersandberg

PR #302, opened 2023-05-09
Fix SSWG incubation badges and update links

alexandersandberg:
Fixed broken SSWG incubation badges and updated their links to point to Swift.org instead of a 404 on GitHub.

### Motivation:

I noticed that the incubation process badges on https://www.swift.org/sswg/incubation-process.html were broken:

CleanShot 2023-05-09 at 09 03 24

Also, the links were outdated and linked to a missing document on GitHub.

### Modifications:

- Fixed the badges by moving the width/height sizing to CSS
- I added a `display: inline-block;` to the badges' styling as well to make them stack horizontally
- I changed the links from `https://github.com/swift-server/sswg/blob/main/process/incubation.md` to `https://swift.org/sswg/incubation-process.html` (we should use an absolute link to Swift.org in case someone reuses them outside Swift.org)

### Result:

This is what it looks like now:

CleanShot 2023-05-09 at 09 04 58

### DIFF
--- a/sswg/incubation-process.md
+++ b/sswg/incubation-process.md
@@ -123,9 +123,9 @@ The SSWG reserves the right to define a singular solution for critical building 
 
 It is recommended for projects that have been accepted to any of the maturity levels to list the maturity level in their project's README with the appropriate badge as defined:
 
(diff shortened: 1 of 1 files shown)


0xTim (approved):
Good catch!

0xTim:
@swift-ci test

PR #304, opened 2023-05-16
improve description of use cases and getting started guides

tomerd:
motivation: address feedback frmo core-team

changes:
* improve description of landing page use cases
* improve description of getting started guides
* add "cross platform command-line tool" as a landing page use case
* change link of use cases to their relevant getting started guides

### DIFF
--- a/index.md
+++ b/index.md
@@ -79,9 +79,10 @@ Swift is a general-purpose programming language built using a modern approach to
     <a href="/getting-started/swiftui" class="cta-secondary">Learn more</a>
   </li>
   <li>
-    <h3>Command-line</h3>
+    <h3>Cross-platform Command-line</h3>
     <p>
-      Writing Swift code is interactive and fun, the syntax is concise yet expressive.
+      Writing Swift is interactive and fun, the syntax is concise yet expressive.
+      Swift code is safe by design and produces software that runs lightning-fast.
       <br><br>
       SwiftArgumentParser and Swift's growing package ecosystem make developing cross-platform command-line tools a breeze.
     </p>
@@ -91,7 +92,6 @@ Swift is a general-purpose programming language built using a modern approach to
   <li>
     <h3>Server and Networking</h3>
     <p>
-      Swift code is safe by design and produces software that runs lightning-fast.
       Swift's small memory footprint, quick startup time, and deterministic performance make it a great choice for server and other networked applications.
       <br><br>
       SwiftNIO and Swift's dynamic server ecosystem bring joy to developing networked applications.
--- a/index.md
+++ b/index.md
@@ -97,7 +97,7 @@ Swift is a general-purpose programming language built using a modern approach to
       SwiftNIO and Swift's dynamic server ecosystem bring joy to developing networked applications.
     </p>
 
-    <a href="/getting-started/vapor-web-server" class="cta-secondary">Learn more</a>
+    <a href="/server" class="cta-secondary">Learn more</a>
   </li>
 </ul>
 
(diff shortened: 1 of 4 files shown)


tomerd:
need help with CSS here since now that we have 3 use cases on the landing page the "boxed" dont quite line up, so either we need a 4th one or we need to fix the CSS so they are aligned

alexandersandberg:
Happy to help with the CSS. I'm not sure having them all on the same line works so well with the longer descriptions:

CleanShot 2023-05-17 at 8 42 05

Here are some alternatives:

1)
CleanShot 2023-05-17 at 8 36 39

2)
CleanShot 2023-05-17 at 8 45 43

3)
CleanShot 2023-05-17 at 8 32 25

I think (1) works quite well. It gives more real estate to "Server and Networking", but it's visually more balanced in regards to the descriptions compared to (2).

index.md
```diff
@@ -72,18 +72,32 @@ Swift is a general-purpose programming language built using a modern approach to
...
     <p>
-      Ever dreamed of using the same programming language on both your mobile applications and their server back-ends? Swift's small footprint, quick startup time, and deterministic performance make it specifically suitable for Server applications.
+      Swift code is safe by design and produces software that runs lightning-fast.
+      Swift's small memory footprint, quick startup time, and deterministic performance make it a great choice for server and other networked applications.
+      <br><br>
+      SwiftNIO and Swift's dynamic server ecosystem bring joy to developing networked applications.
     </p>
 
-    <a href="/server" class="cta-secondary">Learn more</a>
+    <a href="/getting-started/vapor-web-server" class="cta-secondary">Learn more</a>
```
alexandersandberg, line 100: I'm not sure I would link out to the getting started guides here. The use cases on the landing page are meant to give visitors an overview of how Swift can be used for each case, and, IMO, e.g. https://www.swift.org/server/ does a better job with that than the getting started guide. Not sure about the developer.apple.com link though.

(It might be a bit confusing that we're calling this section "Use Cases", and also having a `_use-cases.md` file for the getting started page which is meant for the guides, not "Use cases of Swift". Maybe we can rename the getting started section/file.)

tomerd:
thanks @alexandersandberg - I like option 2 the most

index.md
tomerd: yes, I debated this before making the change. one challange is consistency, ie where should we link the new "command-line tool" use case?

alexandersandberg: I agree. We might need some new content for this actually.

alexandersandberg:
> thanks @alexandersandberg - I like option 2 the most

This is now introduced in [50b7fba](https://github.com/apple/swift-org-website/pull/304/commits/50b7fba01ab705a8c4a0eee11f4861a68894ead4).

index.md
```diff
@@ -72,18 +72,32 @@ Swift is a general-purpose programming language built using a modern approach to
   <li>
     <h3>Apple platforms</h3>
     <p>
-      Swift is a powerful and intuitive programming language on macOS, iOS, watchOS, tvOS and beyond. Swift code is safe by design, yet also produces software that is optimized to run lightning-fast.
+      Swift is a powerful and intuitive programming language optimized when running on iOS, macOS, and other Apple platforms.
+      <br><br>
+      Apple offers a wide variety of frameworks and APIs that make applications developed for these platforms unique and fun.
+    </p>  
+    <a href="/getting-started/swiftui" class="cta-secondary">Learn more</a>
```
tomerd, line 79:
```suggestion
    <a href="https://developer.apple.com/swift/resources/" class="cta-secondary">Learn more</a>
```

alexandersandberg (approved):

PR #324, opened 2023-06-08
remove CLion because JetBrains is no longer supporting Swift

willie:
JetBrains has sunset its Swift support for CLion.
https://plugins.jetbrains.com/plugin/8240-swift

 remove CLion because JetBrains is no longer supporting Swift

### Motivation:

Documentation for CLion Swift support is out of date and can be confusing for people who want to write Swift on other platforms.

### Modifications:

I removed the line referencing CLion.

### Result:

CLion will no longer be on the list of IDEs/Editors with Swift Support


tkremenek:
@swift-ci test

alexandersandberg (approved):

PR #352, opened 2023-08-02
Ignore a few more files to make working with the site a little easier

daveverwer:
This is a non-breaking change that makes this site easier to work with when using Visual Studio Code. This will not break anyone's working environment, whether or not they use Visual Studio Code to edit, or any other editor! It also won't break anything in production.

### Motivation:

When working with the site in Visual Studio Code, it's useful to have a `tasks.json` file so that the Jekyll server starts up and closes down with VS Code. It's also useful to have a per-project `.prettierrc` file for when editors are configured to run prettier on every save, which is setting some people have enabled for front-end work.

### Modifications:

Added several Visual Studio Code per-project configuration files and the `.prettierrc` file to the `.gitignore` so that people who would like to have local copies while developing can easily do so.

I have also added a couple of template files, for those people who would like a basic version of `tasks.json` and an example `.prettierrc`.

### Result:

Nothing changes with this PR. However, it allows people working with the site to have local copies of these files without them getting in the way of other work.


daveverwer:
@swift-ci please test

tomerd (commented):
still more to ignore?

.gitignore
```diff
@@ -18,3 +18,12 @@ xcuserdata
 
 # Build output
 .output
+
+# VS Code files
+.vscode/tasks.json
+.vscode/launch.json
+.vscode/settings.json
+.vscode/extensions.json
```
MaxDesiatov, line 26: Would it be cleaner to ignore the whole `.vscode` directory altogether?

daveverwer: The template file is in there right now. I think Git supports an "Exclude this directory but then explicitly include this file"  if people would prefer it that way?

daveverwer: I think this will work:

```
# VS Code files
.vscode
!.vscode/*.template
```

Pushed a change with that, too.

.prettierrc.template
```diff
@@ -0,0 +1,16 @@
+{
```
tomerd, line 1: including this file intentional?

.vscode/tasks.json.template
```diff
@@ -0,0 +1,19 @@
+{
```
tomerd, line 1: including this file intentional?

.prettierrc.template
alexandersandberg, line 1: Are these config values based on the project's current format, or where are they taken from? I know it's a template, but it's still setting some kind of default for the project.

As formatting configs is a pretty opinionated topic we should be careful with introducing/changing any values.

daveverwer: I included it as Prettier is quite opinionated and doesn’t give you much control like a normal formatter would. It only has the most basic options, and as a template only, I didn’t think it would cause harm.

That said. I have absolutely no appetite to start a conversation about formatting here. All I need is the `.gitignore` change, so I have removed this file.

.prettierrc.template
daveverwer: It was, but see [#r1282370809](https://github.com/apple/swift-org-website/pull/352/files#r1282370809)

.vscode/tasks.json.template
daveverwer: I’ll fight a little for this file, as there’s nothing opinionated in it. It is purely to save someone from needing to look up the syntax to let VS Code start the local server for previews when launching the editor. It doesn’t get in the way of people who don’t use VS Code (as it lives in the hidden `.vscode` directory) and doesn’t affect people who _do_ use it unless they copy it and remove the `.template` extension.

That said. I don’t need it personally. I included it as I thought it might be helpful and will happily remove it.

Better to get rid of it?

alexandersandberg (approved):

.vscode/tasks.json.template
tomerd: I think for now probably better to stay minimalistic

daveverwer: Done

daveverwer:
This is now purely an update to the `.gitignore`. OK to merge?

daveverwer:
@swift-ci please test

tomerd:
👍

PR #361, opened 2023-08-18
Change Windows installation instructions

libewa:
### Motivation:

The Windows installation process errored out on my PC, complaining that `start` could not find the `vs_community.exe` file, which was however present after running `ls`. 

### Modifications:

- Change the installation code block for Windows to not use `start`, but instead execute the installer directly.
- Change the reference to the Microsoft Store from "App Store".

### Result:

The installation instructions for Windows have changed to 
a) not quit with an error
b) correctly reference the Microsoft Store

### DIFF
--- a/download/_windows.md
+++ b/download/_windows.md
@@ -20,7 +20,7 @@ Windows has the following additional platform specific dependencies:
 
 #### Install using the Windows Package Manager
 
-The [Windows Package Manager](https://docs.microsoft.com/windows/package-manager/) can be found in the [App Store](https://www.microsoft.com/en-us/p/app-installer/9nblggh4nns1) or be [installed directly](ms-appinstaller:?source=https://aka.ms/getwinget).
+The [Windows Package Manager](https://docs.microsoft.com/windows/package-manager/) can be found in the [Microsoft Store](https://www.microsoft.com/en-us/p/app-installer/9nblggh4nns1) or be [installed directly](ms-appinstaller:?source=https://aka.ms/getwinget).
 
 #### Install using Scoop
 
--- a/download/_windows.md
+++ b/download/_windows.md
@@ -39,10 +39,7 @@ Invoke-RestMethod -Url 'get.scoop.sh' | Invoke-Expression
    ~~~ cmd
    winget install Git.Git
    winget install Python.Python.3.10
-
-   curl -sOL https://aka.ms/vs/16/release/vs_community.exe
    winget install Microsoft.VisualStudio.2019.Community --force --custom "--add Microsoft.VisualStudio.Component.Windows10SDK.19041 --add Microsoft.VisualStudio.Component.VC.Tools.x86.x64"
-   del /q vs_community.exe
    ~~~
 
    #### With Scoop:
--- a/download/_windows.md
+++ b/download/_windows.md
@@ -36,20 +36,19 @@ Invoke-RestMethod -Url 'get.scoop.sh' | Invoke-Expression
    The platform dependencies cannot be installed through the Windows Package Manager as the install rules do not install the components necessary.  They will be installed through Visual Studio installer.
 
    #### With Winget (Windows Package Manager):
-   ~~~ pwsh
+   ~~~ cmd
    winget install Git.Git
    winget install Python.Python.3.10
 
    curl -sOL https://aka.ms/vs/16/release/vs_community.exe
-   start /w vs_community.exe --passive --wait --norestart --nocache ^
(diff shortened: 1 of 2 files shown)


alexandersandberg:
cc: @compnerd

stevapple:
The Windows installation guide currently handles VS installation manually. We can leverage WinGet features instead:
```
winger install Microsoft.VisualStudio.2019.Community --force --custom "--add Microsoft.VisualStudio.Component.Windows10SDK.19041 --add Microsoft.VisualStudio.Component.VC.Tools.x86.x64"
```

Breakdown:
- The `Microsoft.VisualStudio.2019.Community` installer is exactly the latest version of `https://aka.ms/vs/16/release/vs_community.exe`;
- `--force` ensures that the installer is run even when the latest VS is installed;
- `--custom` passes the supplied flags to the installer, which installs the dependencies for us.

Such approach is much less error-prone as WinGet already counts for different environments (most notably, Powershell or CMD).

stevapple:
Also I feel that installing the community version of VS is a waste because Swift developers don't need the IDE stuff. Visual Studio C++ Build Tools (`vs_buildtools.exe` or `Microsoft.VisualStudio.2019.BuildTools`) is pretty enough.

compnerd:
@libewa did you run from cmd or powershell/pwsh?  The expectation is that you are using the cmd shell.  For automation, the use of `start` is important.

@stevapple yes, winget is the easier solution.  Especially now that winget is a stub on Windows 11 that auto installs.  Perhaps we should swap the order and direct people to the winget instructions?

Also, you are correct that the build tools are sufficient but there are questions around licensing.  VS Community is permitted for use for many cases.  The build tools require a license of Visual Studio and thus the installation of VS.  The instructions seem complicated enough so I think that we should keep to the Visual Studio installation.

libewa:
@compnerd I used pscore7, maybe that's the problem, but I'm going on holiday and can't try any of the solutions.
@stevapple sounds good, as we are installing deps with winget anyways. we should maybe add a note that winget comes preinstalled on win11.

libewa:
also, maybe we should actually use cmd syntax highlighting instead of pwsh

download/_windows.md
```diff
@@ -36,20 +36,19 @@ Invoke-RestMethod -Url 'get.scoop.sh' | Invoke-Expression
...
+   ~~~ cmd
    winget install Git.Git
    winget install Python.Python.3.10
 
    curl -sOL https://aka.ms/vs/16/release/vs_community.exe
-   start /w vs_community.exe --passive --wait --norestart --nocache ^
-     --installPath "%ProgramFiles(x86)%\Microsoft Visual Studio\2019\Community" ^
-     --add Microsoft.VisualStudio.Component.Windows10SDK.19041 ^
-     --add Microsoft.VisualStudio.Component.VC.Tools.x86.x64
+   winget install Microsoft.VisualStudio.2019.Community --force --custom "--add Microsoft.VisualStudio.Component.Windows10SDK.19041 --add Microsoft.VisualStudio.Component.VC.Tools.x86.x64"
```
compnerd, line 44: I wonder if we should just move this to VS2022 at this point and update the SDK to the latest.

getting-started/_installing.md
```diff
@@ -82,32 +82,29 @@ Invoke-RestMethod -Url 'get.scoop.sh' | Invoke-Expression
...
+   ~~~ cmd
    winget install Git.Git
    winget install Python.Python.3.10
 
    curl -sOL https://aka.ms/vs/16/release/vs_community.exe
-   start /w vs_community.exe --passive --wait --norestart --nocache ^
-     --installPath "%ProgramFiles(x86)%\Microsoft Visual Studio\2019\Community" ^
-     --add Microsoft.VisualStudio.Component.Windows10SDK.19041 ^
-     --add Microsoft.VisualStudio.Component.VC.Tools.x86.x64
+   winget install Microsoft.VisualStudio.2019.Community --force --custom "--add Microsoft.VisualStudio.Component.Windows10SDK.19041 --add Microsoft.VisualStudio.Component.VC.Tools.x86.x64"
```
compnerd, line 93: Similar

download/_windows.md
libewa: this is a story for another pr imo

compnerd (approved):
LGTM

alexandersandberg (approved):

alexandersandberg:
@swift-ci please test

alexandersandberg:
@swift-ci please test

PR #370, opened 2023-08-26
Fixed allocations.md markup

### LINKED ISSUE #364: Swift on Server/guides/allocation is totally mangled and missing content
## Describe the bug

The original article https://github.com/swift-server/guides/blob/main/docs/allocations.md was republished as https://www.swift.org/server/guides/allocations.html but unfortunately it doesn't render properly

## To Reproduce

- Try to read the full article
- Try to see the images
- 
## Expected behavior

- Full article readable
- Images available
 
## Actual behavior

- A lot of missing content / images
- A lot seems to be (in unreadable form) under the "Demo program source code" disclosure triangle

natikgadzhi:
This pull request cleans up the markup on https://www.swift.org/server/guides/allocations.html page.

### Motivation:

- Closes #364!

### Modifications:

- To get the collapsible block to work, we would need to add an extra dependency on gh-flavored markdown parser, like this: https://github.com/github/jekyll-commonmark-ghpages — it's not very well maintained it seems, and could have side-effects for other pages.
- So instead, I removed the collapsible details block and replaced it with a sub-subheading.

### Result:

A more readable guide, and hopefully slightly happier @weissi.


server/guides/allocations.md
```diff
@@ -104,17 +104,16 @@ After we have confirmed that our user probe on `malloc` works in general, let's
 
 For example, we could analyse a program which does 10 subsequent HTTP requests using [AsyncHTTPClient](https://github.com/swift-server/async-http-client). If you're interested in the full source code, please expand below.
 
-<details>
-<summary>Demo program source code</summary>
+#### Demo program source code
```
alexandersandberg, line 107: It would be good to use an H3 here to not skip a level.

```suggestion
### Demo program source code
```

alexandersandberg (commented):
Thanks for helping out with this! Seems like a good fix to me.

Left a small comment.

0xTim (approved):
LGTM once @alexandersandberg's comments have been addressed

natikgadzhi:
@alexandersandberg @0xTim thank you both for reviewing "overnight" on a weekend!

server/guides/allocations.md
natikgadzhi: Done!

natikgadzhi:
I also searched the whole repo for other instances of `<details>`, and the only one remaining is commented out.

Gemfile.lock
alexandersandberg, line 1: It looks like this was accidentally committed. Let's remove it. 🙂

alexandersandberg (changes requested):

alexandersandberg:
> I also searched the whole repo for other instances of `<details>`, and the only one remaining is commented out.

There are actually more `<details>` in use, but you have to search for them without the closing bracket as they have a class (`<details class="download">`).

It seems like they are not affected by the same issues though. I think because they all only contain HTML.

natikgadzhi:
Yep. My bad. 
One sec.

Gemfile.lock
natikgadzhi: Cleaned that up. My bad, sorry.

alexandersandberg: Np! 🙂

alexandersandberg (approved):

alexandersandberg:
@swift-ci please test

alexandersandberg:
CI might be down. Will try again later/tomorrow. 😄

alexandersandberg:
@swift-ci test

PR #371, opened 2023-08-28
Add some showcase snippets

bitjammer:


tomerd:
cc @alexandersandberg

alexandersandberg:
Thanks, @bitjammer! Nice to finally have some proper snippets. 🎉 

The snippets have to be placed in the `_data/featured_snippets.yml` file at the moment. I would prefer your way of adding swift files to a folder, but my limited Jekyll knowledge gets in the way here. If anyone else knows how to do this, please feel free to help out!

But for now, I think we have to use the above-mentioned data file. Let me know if you want my help with that.

As the code snippets vary in height quite a bit, we will have to remove the fixed height constraint on the snippets box.

This height constraint was initially added to prevent layout shifts on page reload. Whether that's a real issue we need to fix or not though is up for discussion. This is what it will look like now when reloading:

https://github.com/apple/swift-org-website/assets/35671299/8e3b9b67-8e42-45a2-9296-119fba22e95e

A possible solution would be to only use snippets with roughly the same line length. Then we could add a fixed height and center the snippets within that box. But this also limits us in what we can feature here.

I'm open to suggestions. Maybe the layout shifts aren't a big deal after all, but the longer snippets do push the rest of the content down below the fold, which might not be ideal.

alexandersandberg:
@bitjammer We're planning on merging the content improvements branch very soon. Do you think you will have time to take a look at the feedback above? If not, I can take over. Please let me know! 🙂

alexandersandberg:
I will merge this as is and make the required changes in the content improvements branch. Thanks for providing these snippets! 🙂

bitjammer:
Hi @alexandersandberg, sorry I missed your comments!

I'm not sure what to say about whether the snippet layout container should be variable height from a layout perspective. However, being text, the snippet *content* should be allowed to variable in size so we don't limit our expressiveness here. I can totally see that there are limits to that, though! We wouldn't want them to be too long and generally speaking snippets are encouraged to be as brief as possible. I admit that's not really a precise answer.

Just looking at your video, I don't know if it's as much of a problem that the layout differs slightly on _reload_. If things were resizing with a dynamic UI, like with a prev/next button, I'd be a bit more concerned.

PR #373, opened 2023-08-29
Content improvement guide updates

0xTim:
Updates the guides so they match what's generated for 5.8 to avoid confusion

cc @tomerd


tomerd (approved):
thanks

PR #377, opened 2023-09-05
Add value and reference types article

dempseyatgithub:
- Update link from Go Further section of Getting Started page
- Add a bit of CSS to display author info correctly.

Should probably move to an ‘article’ layout that encapsulates this.

### DIFF
--- a/documentation/articles/value-and-reference-types.md
+++ b/documentation/articles/value-and-reference-types.md
@@ -1,0 +1,182 @@
+---
+layout: page
+date: 2023-04-29 12:00:00
+title: Value And Reference Types In Swift
+author: [jamesdempsey]
+---
+
+Types in Swift are grouped in two categories: _value types_ and _reference types_. Each behave differently and understanding the difference is an important part of understanding Swift.
+
+If you are new to programming or coming to Swift from another language, these concepts may be new to you.
+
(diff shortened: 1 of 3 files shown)


dempseyatgithub:
FYI, I am traveling today - so will have limited access to Github / laptop during the day.

dempseyatgithub:
Note that since this is a single article, I only linked to it from the Go Further section, which is sufficient to launch the content improvements.

I think we should add at least a couple more articles before we add an Articles section to the documentation page. It looks very empty to have a single entry.

tomerd (approved):

documentation/articles/value-and-reference-types.md
```diff
@@ -0,0 +1,182 @@
...
+    font-size: 1.75rem;
+    margin-top: .5em;
+    color: crimson;
+    font-weight: 600;
+  }
+}
+</style>
+
+<hr>
+Contributed by
```
tomerd, line 152: could be nice to extract this to a snippet that can be included in similar articles

dempseyatgithub: Yes, my thought was we’d add an article layout that would encapsulate this.

It may also make sense to move the byline styles into classes that aren’t dependent on containment.

I just submitted Issue #378 to cover this.

dempseyatgithub:
@swift-ci test

PR #379, opened 2023-09-05
Resolve issue with Red Hat download links

kaishin:
This PR proposes a solution to links for Red Hats download links being broken due to a mismatch between the logic constructing the URL and the actual URL.


shahmishal (approved):

shahmishal:
Thanks!

PR #401, opened 2023-09-17
Fix download footnotes links

natikgadzhi:
Fixes the links in downloads > older releases blocks in `download/_build-platform.html`. The links used markdown syntax in a html file, and hence would not render correctly.

The first block of footnotes on the page rendered correctly, only the "Older releases" were broken. I've only spotted them because some of the Xcode download links were broken.

After/before:
Screenshot 2023-09-17 at 12 26 04 PM

### DIFF
--- a/download/_build-platform.html
+++ b/download/_build-platform.html
@@ -67,5 +67,5 @@
     </tbody>
 </table>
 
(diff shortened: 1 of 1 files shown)


shahmishal:
@swift-ci test

shahmishal (approved):

PR #408, opened 2023-09-20
Prefer linking to GitHub profiles in blog post bylines

dempseyatgithub:
This PR reverses the preference order of the link used in blog post bylines.

This change will link to the author’s GitHub profile if present, then fall back to link to the Twitter/X profile if present, finally displaying no link if neither is present.

Twitter/X now requires being logged-in to view profile information. GitHub displays profile information without being logged-in. This change provides a better experience by not requiring a user to be logged into any service to view the profile information linked from a blog post byline.

The change is minor, changing the order in which presence of the account info is checked. I’ve tested this on a local copy of the site.

### DIFF
--- a/_layouts/post.html
+++ b/_layouts/post.html
@@ -18,10 +18,10 @@ layout: base
         {% endif %}
 
         <span class="author">
-          {% if author.twitter %}
-            <a href="https://twitter.com/{{ author.twitter }}/" rel="nofollow" title="{{ author.name }} (@{{ author.twitter}}) on Twitter">{{ author.name }}</a>
-          {% elsif author.github %}
+          {% if author.github %}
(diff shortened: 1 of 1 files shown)


alexandersandberg (approved):
I think this is a good idea. We could also consider adding the option to link to the author's Swift forums account, but I'm not sure if we have any data for that yet.

dempseyatgithub:
@swift-ci test

dempseyatgithub:
> We could also consider adding the option to link to the author's Swift forums account, but I'm not sure if we have any data for that yet.

We currently have GitHub info for all authors except for three. We would need to gather Swift forum info for all authors.

One reason I think GitHub profile pages are preferable is that they have dedicated fields for users to provide various means of contact or connection: URL, email, multiple social media handles. The Swift Forums profile only provides a field for a URL.

I would also guess that developers are more likely to curate their GitHub profile than their Swift forum profile, but that’s just a guess.

0xTim (approved):
Big +1 to this - feel free to merge!

PR #417, opened 2023-10-01
[Windows] Fixes version of the Python needed for LLDB to working properly

HeMet:
### Motivation:

It seems, Installation Guide for Windows has incorrect version of Python specified.
See https://forums.swift.org/t/unable-to-debug-on-windows/67609

### Modifications:

I've changed version of Python required from `3.10` to `3.9`

### Result:

The issue with debugger is solved.

### DIFF
--- a/install/windows/_winget.md
+++ b/install/windows/_winget.md
@@ -6,7 +6,7 @@
 
    ~~~ batch
    winget install Git.Git
-   winget install Python.Python.3.10
+   winget install Python.Python.3.9
    winget install Microsoft.VisualStudio.2019.Community —force —custom "—add Microsoft.VisualStudio.Component.Windows10SDK.19041 —add Microsoft.VisualStudio.Component.VC.Tools.x86.x64"
    ~~~
 


install/windows/_winget.md
```diff
@@ -6,7 +6,7 @@
 
    ~~~ batch
    winget install Git.Git
-   winget install Python.Python.3.10
+   winget install Python.Python.3.9
```
compnerd, line 9: I think that we should externalise this to a table.  The different releases of Swift require different versions.  If I'm not mistaken, 5.9.0 requires 3.11 🤦

compnerd: Okay, verified that the 5.9 release binaries do need python39.dll.

compnerd:
@shahmishal ping

compnerd:
@swift-ci please test

compnerd:
@shahmishal @tomerd - can we get this merged?

shahmishal (approved):

PR #426, opened 2023-10-19
[Download] Release Swift 5.9.1

shahmishal:
Screenshot 2023-10-18 at 6 15 12 PM

### DIFF
--- a/_data/builds/swift_releases.yml
+++ b/_data/builds/swift_releases.yml
@@ -1335,26 +1335,26 @@
       archs:
         - x86_64
     - name: Ubuntu 20.04
-      docker: 5.9.1-focal
+      docker: Coming Soon
       archs:
         - x86_64
         - aarch64
     - name: Ubuntu 22.04
-      docker: 5.9.1-jammy
+      docker: Coming Soon
       archs:
         - x86_64
         - aarch64
     - name: CentOS 7
-      docker: 5.9.1-centos7
+      docker: Coming Soon
       archs:
         - x86_64
     - name: Amazon Linux 2
-      docker: 5.9.1-amazonlinux2
+      docker: Coming Soon
       archs:
         - x86_64
         - aarch64
     - name: Red Hat Universal Base Image 9
-      docker: 5.9.1-rhel-ubi9
+      docker: Coming Soon
       dir: ubi9
       archs:
         - x86_64
--- a/_data/builds/swift_releases.yml
+++ b/_data/builds/swift_releases.yml
@@ -1329,7 +1329,7 @@
   tag: swift-5.9.1-RELEASE
   xcode: Xcode 15
   xcode_release: false
-  date: 2023-10-18
+  date: 2023-10-19
   platforms:
     - name: Ubuntu 18.04
       archs:
--- a/_data/builds/swift_releases.yml
+++ b/_data/builds/swift_releases.yml
@@ -1335,26 +1335,26 @@
       archs:
         - x86_64
     - name: Ubuntu 20.04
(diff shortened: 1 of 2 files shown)


shahmishal:
@swift-ci test

alexandersandberg (approved):

cthielen (approved):

download/_build-platform.html
```diff
@@ -66,6 +66,3 @@
         {% endfor %}
     </tbody>
 </table>
-
-<sup>1</sup> Swift {{ include.platform.name }} is available as part of <a href="https://itunes.apple.com/app/xcode/id497799835">{{ include.platform.xcode }}</a>.<br>
```
stevapple, line 70: Why is this line removed?

download/_build-platform.html
```diff
@@ -66,6 +66,3 @@
...
-<sup>2</sup> Swift {{ include.platform.name }} Windows 10 toolchain is provided by <a href="https://github.com/compnerd">Saleem Abdulrasool</a>. Saleem is the platform champion for the Windows port of Swift and this is an official build from the Swift project. <br><br>
```
stevapple, line 71: We can add this one conditionally according to Swift version. I don’t think it should be dropped for existing releases.

download/_build-platform.html
```diff
@@ -52,7 +50,9 @@
                 {% endif %}
                 {% assign tag_downcase = include.platform.tag | downcase %}
                 <a href="https://download.swift.org/{{ tag_downcase }}/{{ platform_name_url }}/{{ include.platform.tag}}/{{ include.platform.tag}}-{{ platform_name }}.{{ package_extension }}">{{ arch }}</a>
+                {% if platform.name != "Windows 10" %}
```
stevapple, line 53: We’re still documenting how to verify the signature of the Windows toolchain, which works pretty well (and is depended by some CI actions IFAIK). Can you expand on why we’re hiding this for Windows?

_data/builds/swift_releases.yml
```diff
@@ -1325,3 +1325,40 @@
...
+      archs:
+        - x86_64
+        - aarch64
+    - name: Red Hat Universal Base Image 9
+      docker: 5.9.1-rhel-ubi9
+      dir: ubi9
+      archs:
+        - x86_64
+        - aarch64
+    - name: Windows 10
```
stevapple, line 1362: Since Microsoft has released Windows 11 a pretty long while ago, and it’s binary compatible with Windows 10, it makes more sense to change this into "Windows 10/11", even if it’s `windows10` internally. Adding 11 here would make it more intuitive for new users.

stevapple (commented):
Sorry for being late in this thread. I noticed a few regressions from the "Download" page after 5.9.1 release (most of which are Windows-related), so I may want to ask here first.

_data/builds/swift_releases.yml
shahmishal: @compnerd

download/_build-platform.html
shahmishal: I will get this updated

download/_build-platform.html
shahmishal: good point, we should add Swift 5.9.0 or older check.

download/_build-platform.html
shahmishal: We need to fix this statement, as it's not true with all of the releases.

download/_build-platform.html
stevapple: Implemented in #429

PR #429, opened 2023-10-24
Refactor "Download" page, clean up unused codes, and update Windows toolchain information

stevapple:
This PR refined the structure of "Download" page with cleanups and information updates.

### Motivation:

Some Windows toolchain information is either wrong, outdated or accidentally removed by prior patches. This PR tries to fix them as well as cleans up more codes in the "Download" page.

### Modifications:

- Removes unused templates in `download/`;
- Removes unused variables in `download/index.md`;
- Update some template and argument names in `download/`;
- Update Windows toolchain provider;
- Add conditional information of Windows toolchain provider;
- Fix wrong information about Windows platform support;
- Add `platform` metadata to `swift_releases.yml`;
- Other trivial fixes.

### Result:

_There used to be an image, but it is too large for some devices and would prevent contributors from browsing the thread._

_In short, it can build and the result is much like the one before #426, except for the removed Xcode provider notice._


stevapple:
@compnerd You may also want to check out this one since it's related to the toolchain (and CI?) provider change.

@shahmishal May you review this if you have time?

shahmishal (approved):
Thank you @stevapple! LGTM

shahmishal:
@swift-ci test

platform-support/_platform-support.md
```diff
@@ -81,7 +81,7 @@ Platform owners help facilitate contributors to the Swift project to collaborate
 |:--------------------:|:-----------------------------------:|:-------------------------------------------------:|:----------------------------:|
 | **Apple platforms**  | [Docs](/getting-started/#on-macos)  | [Apple Inc.](https://www.apple.com)               | ✓                            |
 | **Linux**            | [Docs](/getting-started/#on-linux)  | [Apple Inc.](https://www.apple.com)               | ✓                            |
-| **Windows**          | [Docs](/getting-started/#on-windows)| [Saleem Abdulrasool](https://github.com/compnerd) | ✓                            |
+| **Windows**          | [Docs](/getting-started/#on-windows)| [Apple Inc.](https://www.apple.com)               | ✓                            |
```
finagolfin, line 84: What is the significance of this change, is Saleem stepping down as maintainer?

shahmishal: Saleem (cc: @compnerd) is not stepping down from my understanding, @stevapple might have changed this because Apple is now producing and code signing the Windows toolchain compared to Saleem.

stevapple: Yep, I was trying to reflect the fact that Swift on Windows toolchains are now produced, signed and provided by Apple. This does not indicate any changes to platform maintainers, and as you can see Saleem remain to be the release manager of Windows platform for upcoming releases.

compnerd: This page is about the downloads, which are now signed by Apple as they are being built on swift.org now 🎉.  I suppose that there may be some value in tracking the platform maintainers, but I think that the work that is going on to create a Platform Steering Group should address that need.

Issue #428, opened 2023-10-24
Properly document historic dependency compatibility and feature set of the Windows toolchain


stevapple:
## Is your feature request related to a problem? Please describe

The Windows toolchain is evolving so fast that the compatibility status and features may change in each release. Currently Swift.org only documents the latest release, and the legacy part is incomplete.

## Describe the solution you'd like

We should:
- state clearly that some information only applies to the latest version;
- hide unnecessary information according to user-selected toolchain version;
- check and complete the information table for historic releases.

The UI/UX can be achieved by a dynamic version selector (and route through hash history?), or let the generator to generate static pages for all releases.

## Describe alternatives you've considered

We can continue to use a single page, but the page would be messy and hard to read.

## Additional context

Linux toolchain dependencies and macOS version compatibility are also changing over time. If there's a way to handle it, we should also apply to other platforms.

PR #431, opened 2023-10-26
Revise WinGet and Scoop installation guide

stevapple:
Revise the WinGet and Scoop installation guide to make them clearer and more robust.

### Motivation:

Due to different CI environments being used, different Swift on Windows releases depend on different minor releases of Python. I’ve managed to update all WinGet manifests based on the [analysis](https://github.com/stevapple/swift-windows-toolchain-analysis), and we should drop manual dependency installation step and let WinGet decide what dependencies to install instead.

### Modifications:

In `install/_winget.md`:
- No longer suggest users to manually install Git and Python;
- No longer require installing Python module `six`;
- Clarify the Visual Studio install step;
- Add a note about Visual Studio edition choosing.

In `install/_scoop.md`:
- No longer require installing Python module `six`;
- Add a note about Visual Studio edition choosing;
- Use Python 3.9 instead of 3.10 to match Swift 5.8.1+.

### Result:

- The WinGet installation guide now applies to all Swift on Windows releases (published in WinGet).
- The Scoop installation guide now works for Swift 5.8.1+.

### DIFF
--- a/install/windows/_winget.md
+++ b/install/windows/_winget.md
@@ -17,11 +17,3 @@
    ~~~ batch
    winget install --id Swift.Toolchain -e
    ~~~
-
-0. Install required Python libraries:
-
-   Start up a new command-line shell (eg. Command Prompt, Windows PowerShell) and install the Python library `six`.
-
-   ~~~ batch
-   pip install six
-   ~~~
--- a/install/windows/_winget.md
+++ b/install/windows/_winget.md
@@ -2,22 +2,26 @@
 
 [Windows Package Manager](https://docs.microsoft.com/windows/package-manager/) (aka WinGet) comes pre-installed with Windows 11 (21H2 and later). It can also be found in the [Microsoft Store](https://www.microsoft.com/p/app-installer/9nblggh4nns1) or be [installed directly](ms-appinstaller:?source=https://aka.ms/getwinget).
 
-0. Install required dependencies:
+0. Install required Visual Studio components:
+
+   Install the latest MSVC toolset and Windows 11 SDK (10.0.22000) through Visual Studio 2022 Community installer. You may change the Visual Studio edition depending on your usage and team size.
 
    ~~~ batch
-   winget install --id Git.Git -e
-   winget install --id Python.Python.3.9 -e
    winget install --id Microsoft.VisualStudio.2022.Community --exact --force --custom "--add Microsoft.VisualStudio.Component.Windows11SDK.22000 --add Microsoft.VisualStudio.Component.VC.Tools.x86.x64"
    ~~~
+
+0. Install Swift and other dependencies:
+
+   Install the latest Swift developer package, as well as compatible Git and Python tools if they don't exist.
+
+   ~~~ batch
+   winget install --id Swift.Toolchain -e
+   ~~~
+
+0. Install required Python libraries:
 
    Start up a new command-line shell (eg. Command Prompt, Windows PowerShell) and install the Python library `six`.
 
    ~~~ batch
    pip install six
-   ~~~
-
-0. Install Swift:
-
-   ~~~ batch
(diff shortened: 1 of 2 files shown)


stevapple:
@compnerd The remaining problem is that the `pip` in system path is not necessarily the one used by LLDB. Any trick to detect and guide users to install `six` in a robust way?

install/windows/_winget.md
```diff
@@ -2,22 +2,26 @@
...
    ~~~ batch
-   winget install --id Git.Git -e
-   winget install --id Python.Python.3.9 -e
    winget install --id Microsoft.VisualStudio.2022.Community --exact --force --custom "--add Microsoft.VisualStudio.Component.Windows11SDK.22000 --add Microsoft.VisualStudio.Component.VC.Tools.x86.x64"
    ~~~
 
-   Start up a new command-line shell (eg. Command Prompt, Windows PowerShell) and install the Python library `six`.
+0. Install Swift and other dependencies:
+
+   Install the latest Swift developer package, as well as compatible Git and Python tools if they didn't exist.
```
compnerd, line 15:
```suggestion
   Install the latest Swift developer package, as well as compatible Git and Python tools if they don't exist.
```

compnerd:
> @compnerd The remaining problem is that the `pip` in system path is not necessarily the one used by LLDB. Any trick to detect and guide users to install `six` in a robust way?

Unfortunately, that isn't so simple. The way that `python.dll` is found is via `path`. So whatever is resolved as `python` is _likely_ the right one (it is technically possible to have `Python.dll` installed without the interpreter, and have that be given higher precedence, which would cause the incorrect resolution).

Assuming that they have the interpreter installed as well, the trick that comes to mind is:

```sh
python -m pip install --user six
```

stevapple:
@compnerd The problem is LLDB depends on specific minor version by loading, eg. `python310.dll`, while Python installation doesn’t expose something like `python3.10.exe` that we can run directly (and it’s also not a simple task to get the version itself).

stevapple:
Okay so it turns out that LLDB doesn't require it any more and Swift 5.8+ has adopted the change. I wonder if it's really required in the history because LLDB always vendors its own copy (maybe it's just misconfigured so the copy cannot be detected?).

Anyway, decide to remove the step now🎉

compnerd:
> @compnerd The problem is LLDB depends on specific minor version by loading, eg. `python310.dll`, while Python installation doesn’t expose something like `python3.10.exe` that we can run directly (and it’s also not a simple task to get the version itself).

I don't think that we have any way to really control the library name (and it impacts the distribution of Python itself). The dependency on python is rather frustrating because we cannot use static linking as it prevents the use of lldb for scripted operations.

Could we perhaps use some horrible PowerShell script to determine the version and then find the install by scanning the registry?  It would be pretty unsightly, but, I don't see a good solution here. We _could_ automate some of this with a custom action that executes optionally, but I really would prefer avoiding the additional code for this.

compnerd:
> Anyway, decide to remove the step now🎉

This sounds awesome!

Thanks for your help with the installation instructions.

install/windows/_winget.md
stevapple: fixed👌

compnerd (approved):
Probably should clear up the summary to remove the "move `six` to the end" bit.

stevapple:
Anyone to review, perhaps @tomerd or @shahmishal?

shahmishal (approved):

shahmishal:
@swift-ci test

shahmishal:
Thank you @stevapple!

stevapple:
Ping for merge.

Also want to know when SWWG will hold the next meeting, in which I would appreciate if you can discuss #428.

stevapple:
Ping

PR #443, opened 2023-11-09
Add docs on how to wrap C/C++ library in Swift

yim-lee:
Co-authors: @etcwilde, @ktoso, @yim-lee


_data/go_further.yml
```diff
@@ -26,4 +26,11 @@
   thumbnail_url: /assets/images/getting-started/regular-expression-thumbnail.jpg
   release_date: 2022-06-07
   external: true
+- title: "Wrapping C/C++ Library in a Swift Package"
+  description: "This article describes how to wrap a C or C++ library in a Swift package."
```
etcwilde, line 30: This is a bit more general than just packages, isn't it? It has examples of wrapping C for Swift packages and for CMake.

yim-lee: Changed title to "Wrapping C/C++ Library in Swift" https://github.com/apple/swift-org-website/pull/443/commits/40be479bc1ae107d770205cd9258511afbb53d26

yim-lee:
@swift-ci test

yim-lee:
cc @Lukasa @FranzBusch @weissi

documentation/articles/wrapping-c-cpp-library-in-swift.md
```diff
@@ -0,0 +1,397 @@
...
+
+    deinit {
+        rocksdb_writeoptions_destroy(underlying)
+    }
+}
+```
+
+The downside of non-copyable types is that currently they cannot be used in all contexts. For example in Swift 5.9 it is not possible to store a non-copyable type as a field, or pass them through closures (as the closure could be used many times, which would break the uniqueness that a non-copyable type needs to guarantee). The upside is that, unlike classes, no reference counting is performed on non-copyable types.
+
+
```
ktoso, line 350: I'll add some information about Strings as well.
What else should we cover here @Lukasa @FranzBusch @weissi ?

Lukasa: A few thoughts:

1. Extend the above to note that it is easy to accidentally defeat these helpers by accessing the `underlying` field from outside the method. Make it `private` and only access it from methods within the class, including using `with` scoped functions if needed.
2. Talk about offering safe abstractions as quickly as possible. Don't use `WriteOptions` as a pointer wrapper, encapsulate it, e.g. https://github.com/apple/swift-crypto/blob/f4b21dbde7fd2b91d7f65e0b80f78c07dea1abc7/Sources/Crypto/Util/BoringSSL/ArbitraryPrecisionInteger_boring.swift#L173
3. Talk about adding value semantics where appropriate, again `ArbitraryPrecisionInteger` is a good example (https://github.com/apple/swift-crypto/blob/f4b21dbde7fd2b91d7f65e0b80f78c07dea1abc7/Sources/Crypto/Util/BoringSSL/ArbitraryPrecisionInteger_boring.swift#L60).

documentation/articles/wrapping-c-cpp-library-in-swift.md
```diff
@@ -0,0 +1,397 @@
+---
+layout: page
+date: 2023-11-08 12:00:00
+title: Wrapping C/C++ Library in Swift
+author: [etcwilde, ktoso, yim-lee]
+---
+
+## Package
```
FranzBusch, line 8: I think this article could use a bit of an introduction that covers the overall problem it tries to explain and what the example is going to be focused on

documentation/articles/wrapping-c-cpp-library-in-swift.md
```diff
@@ -0,0 +1,397 @@
...
+      	// Relative paths under 'CMyLib' of the source 
+      	// files and/or directories. For example:
+      	// "./my-lib/src/foo.c",
+      	// "./my-lib/src/baz",
+    	],
+    	cSettings: [
+      	// .headerSearchPath("./my-lib/src"),
+    	]
+	),
+	```
```
FranzBusch, line 44: The formatting of the code blocks here look off

documentation/articles/wrapping-c-cpp-library-in-swift.md
FranzBusch: I agree with Cory's points here. We should really push developers to safely wrap the underlying types as fast as possible into safe construct and at best contain the unsafeness to a special place in their wrapping code. Above that nothing should see the unsafe construct. Additionally, I am missing a bit about the various pointer bridges that one has to go through. I know that I often had to write C code myself to properly type the pointers so that I get correct imports in Swift. Lastly, I would love a short part about `GNU_SOURCE` which often comes up as well.

documentation/articles/wrapping-c-cpp-library-in-swift.md
yim-lee: amended in https://github.com/apple/swift-org-website/pull/443/commits/86412f25b1032924c9cdff5747c70e2c7966e132

documentation/articles/wrapping-c-cpp-library-in-swift.md
yim-lee: added introduction in https://github.com/apple/swift-org-website/pull/443/commits/737c8a42602e4a8842a5c314cf51b907297f3091

documentation/articles/wrapping-c-cpp-library-in-swift.md
ktoso: > Lastly, I would love a short part about GNU_SOURCE which often comes up as well.
I don't know about that, maybe I could ask you to add a few words? :)

FranzBusch: So my understanding is that if you define `GNU_SOURCE` you will:

- Gain access to lots of non-standard GNU extension functions
- Gain access to functions which were not part of the POSIX standard
- Gain access to low-level functions that cannot be portable

The other interesting bit around `GNU_SOURCE` is that you must use the define in an implementation file and not a header file otherwise stuff will go wrong. @Lukasa and @weissi probably know the details of `GNU_SOURCE` better than me though

tomerd (approved):

tomerd:
@swift-ci test

PR #466, opened 2023-12-20
Add Swift 5.9.2 to evolution filter buttons

dempseyatgithub:
Proposal SE-0407 was implemented in Swift 5.9.2.

Adding version 5.9.2 to the swift evolution script so that filter buttons continue to work for all implemented proposals.


dempseyatgithub:
@swift-ci test

0xTim (approved):

Issue #467, opened 2023-12-20
Invalid version in evolution dashboard query string causes inconsistent state


dempseyatgithub:
Use a URL to access the evolution dashboard with an invalid query string.

For example:

`https://www.swift.org/swift-evolution/#?version=5.5.6`

The filter indicator is active.

Clicking the 'Show Filters' tab displays all of the proposal status and version filter buttons. The 'Implemented' status is not selected.

Clicking the 'Implemented' filter button hides the version buttons.

What should happen is if an invalid version is in the query string, it should have no effect on the filter status of the dashboard.

dempseyatgithub:
It looks like the javascript uses the original `version` and `status` parameters to check if filtering interface elements should be turned on. It should use the array of processed parameters which have invalid items removed.

PR #470, opened 2023-12-21
Fix issue 467

dempseyatgithub:
This PR addresses Issue #467 

Without this fix, an invalid version or status in the fragment query of an evolution dashboard URL would display the filter interface. For an invalid version, it also gets the interface in a state where selecting the Implemented status hides the version filters.

Prior to this change, the script was checking the raw parameters in the query string rather than the array of validated values.

Finally this PR updates the comment that describes the fragment query syntax.


dempseyatgithub:
@swift-ci test

dempseyatgithub:
@swift-ci test

0xTim (approved):

PR #495, opened 2024-01-25
Use 'current version' instead of specific Xcode version

### LINKED ISSUE #493: Install page references outdated Xcode version as latest
From feedback on the top navigation changes on the forums: https://forums.swift.org/t/navigation-improvements-to-swift-org/69586/6

The install page at https://www.swift.org/install/ has the text:

"Download Xcode 15.1 which contains the latest Swift release."

But Xcode 15.2 has been released.

It probably isn't necessary to refer to the exact version of Xcode on the page, since the link to the Mac App Store will note the current version.

Would probably make sense to reword to something like:
"Download the current version of Xcode which contains the latest Swift release."

dempseyatgithub:
The Xcode version is read from the last entry of swift_releases.yml. Multiple Xcode versions can ship with the same Swift version, so the Xcode version displayed can be out of date.

The exact Xcode version is not very important to display on the page. So, instead of creating and maintaining an Xcode version data file, the text changes to say 'current version'.

Since the user will clicking through to the Mac App Store, they will see the correct version number on the app page in the store.

This fixes #493.

### DIFF
--- a/install/macos/_macos.md
+++ b/install/macos/_macos.md
@@ -4,7 +4,7 @@
   <li class="resource featured">
     <h3>Xcode</h3>
     <p class="description">
-       Download {{ site.data.builds.swift_releases.last.xcode }} which contains the latest Swift release. 
+       Download the current version of Xcode which contains the latest Swift release. 
     </p>
     <a href="https://itunes.apple.com/app/xcode/id497799835" class="cta-secondary">Download Xcode</a>
   </li>
--- a/install/macos/index.md
+++ b/install/macos/index.md
@@ -7,7 +7,7 @@ title: macOS Installation Options
 
 ## Installation via Xcode
 
-Download [{{ site.data.builds.swift_releases.last.xcode }}](https://itunes.apple.com/app/xcode/id497799835) which contains the latest Swift release. 
+Download the [current version of Xcode](https://itunes.apple.com/app/xcode/id497799835) which contains the latest Swift release. 
 
 ## Installation via Swift.org package installer
 


kaishin (approved):
+1 from me

dempseyatgithub:
Note I also made the change to the macOS install page where a similar thing could happen.

Will defer merging to give @shahmishal a chance to take a look.

dempseyatgithub:
@swift-ci test please

alexandersandberg (commented):
If the `xcode` values are inaccurate we should probably remove them from the data file as well.

install/macos/_macos.md
```diff
@@ -4,7 +4,7 @@
   <li class="resource featured">
     <h3>Xcode</h3>
     <p class="description">
-       Download {{ site.data.builds.swift_releases.last.xcode }} which contains the latest Swift release.
```
shahmishal, line 7: What about `Download {{ site.data.builds.swift_releases.last.xcode }} or later which contains the latest Swift release`

In this example: Xcode 15.1 and 15.2 both have same swift compiler, so we would cover the case where we don't release a newer swift compiler version with a Xcode release.

shahmishal: We can keep it to Xcode 15.1, however the text would be 

```
Download {{ site.data.builds.swift_releases.last.xcode }} or later contains the latest Swift release
```
output:
Download Xcode 15.1 or later contains the latest Swift release

dempseyatgithub: Thank you for clarifying, I missed the 'or later' in your original post.

So the choice would be between:

"Download Xcode 15.1 or later which contains the latest Swift release."

"Download the current version of Xcode which contains the latest Swift release."

In both cases, the user doesn't know exactly which version of Xcode they are going to get when they click on the link to the Mac App Store.

I think the more general 'current version' reads better and also gives the user a better conceptual idea of what they are doing.

It's probably less important to someone new to Swift what the exact version of Xcode is that they have to download. But probably more important that they are getting the current version of Xcode to get the latest version of Swift.

shahmishal: Good point, I am ok with using current compared to specific Xcode version.

shahmishal (approved):

PR #513, opened 2024-02-05
Project ideas page for GSoC 2024

ktoso:
This is necessary for our GSoC 2024 application so we need to get this merged right now, even if still empty.

It is the same shape of page and content as we did for all previous years -- could I request a quick merge here and we can then follow up with adding more content? @alexandersandberg @0xTim et al?

### DIFF
--- a/gsoc2024/index.md
+++ b/gsoc2024/index.md
@@ -1,0 +1,55 @@
+---
+layout: page
+title: Project Ideas for GSoC 2024
+---
+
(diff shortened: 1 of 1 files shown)


alexandersandberg (approved):
Sounds good! 🙂

ktoso:
Thank you! Please feel free to review folks -- and I'd follow up on any comments in a separate PR

PR #523, opened 2024-02-09
February’s package page updates.

daveverwer:
This update contains the February data file with new packages for the [Packages page](https://www.swift.org/packages/).

Packages in the Community Showcase were nominated in [this thread on the Swift forums](https://forums.swift.org/t/nominations-for-the-packages-community-showcase-on-swift-org/68168) and voted on by the [SWWG](https://www.swift.org/website-workgroup/).


daveverwer:
@swift-ci please test

_data/packages/packages.yml
```diff
@@ -401,16 +390,30 @@ categories:
...
+          developers more confidence with less code.
+        owner: Apple
+        swift_compatibility: 5.9+
+        platform_compatibility:
+          - Apple
+          - Linux
+        platform_compatibility_tooltip: Apple (iOS, macOS, visionOS, watchOS, tvOS) and
+          Linux
+        license: Apache 2.0
+        url: https://swiftpackageindex.com/apple/swift-testing
```
MahdiBM, line 416: This is not quite true, swift-testing doesn't support Swift 5.9 or even 5.10 and relies on some new integrations available only in the main branch (Swift 5.11 hopefully? Already awaiting the WWDC video 😁).
Considering swift-testing is in such an active development and doesn't support the widely-used Swift versions, it might be better to wait a while before including it here 🤔

_data/packages/packages.yml
```diff
@@ -9,82 +9,81 @@ categories:
...
-        url: https://swiftpackageindex.com/space-code/typhoon
-        note: Discussed on [Episode 37 of Swift Package Indexing](https://share.transistor.fm/s/de52cb62).
-      - name: SwiftUICoreImage
-        description: Provides a convenient way to chain multiple Core Image filters to
-          CIImage instances and render them into SwiftUI or any other context. It can
-          also be used without SwiftUI.
-        owner: Dan Wood
-        swift_compatibility: 5.7+
+        platform_compatibility_tooltip: Apple (iOS, macOS, visionOS, watchOS, tvOS)
+        license: Unknown license
```
kaishin, line 73: The license is [MIT](https://github.com/twostraws/Vortex?tab=License-1-ov-file#readme).

daveverwer: This is coming through as GitHub can’t detect the license on Vortex, which is where the Swift Package Index gets its license identification from. See, for example, another repository of Paul’s has been correctly identified:

Screenshot 2024-02-09 at 17 45 14@2x

whereas this one shows “License”, meaning _something_ in there has been tweaked from the standard MIT license:

Screenshot 2024-02-09 at 17 47 47@2x

In this instance, it could be as simple as the addition of the word “Vortex” at the top, but the GitHub identification is deliberately strict so as not to accidentally mislabel things as a specific license when they are not 100% sure.

This is not something we are planning to change with Swift Package Index, and I’d suggest we don’t change it here either. If people click through to the license from the Package Index page, they can read it for themselves and decide whether to use the package.

_data/packages/packages.yml
daveverwer: I’ve not tested (no pun intended!) this package myself, but the reason we’re showing 5.9 compatibility is because the 0.3.0 version does build [successfully with Swift 5.9 across all platforms](https://swiftpackageindex.com/apple/swift-testing/builds). `main` no longer does, but the compatibility brought into this page on Swift.org is a simplified/condensed version.

If we decide not to include this package on Swift.org, I’d need to make some changes to the software that generates this data file to incorporate a deny/exclude list as we don’t intend to remove this package from the Swift Package Index, and it is ranking high for the “testing” keyword, which is where this data is coming from. I’ll take the advice of the rest of the workgroup on whether they agree this should not be listed.

The Swift Package Index doesn’t yet check support for 5.10 compatibility, but it will soon-ish (that’s as specific as I can be, unfortunately, as we have a lot on at the moment).

0xTim: I think it should still be included as it was nominated and voted for. We can discuss specifics in the next workgroup meeting but the aim of the package nomination is to showcase interesting packages and the wide range of packages, not necessarily drive adoption IMO

_data/packages/packages.yml
daveverwer: I've raised a PR on The Vortex repository that would fix this:

https://github.com/twostraws/Vortex/pull/5

_data/packages/packages.yml
daveverwer: > as it was nominated and voted for

Interestingly, this one was not nominated or voted for. This is in the ["Testing" category](https://www.swift.org/packages/testing.html) which is powered by the SPI ranking algorithm for packages tagged with "testing". It's now worked its way up to be in the top testing packages, and so is now included here.

MahdiBM: I don't have any strong opinions about this but i'm just saying that considering you're currently avoiding including previously-included packages, there might be a better time for swift-testing to be included.

See [this post](https://forums.swift.org/t/package-swift-testing-is-using-swift-tools-version-5-10-0/69821) for more "official" info about swift-testing's situation.

daveverwer: > I don't have any strong opinions about this but i'm just saying that considering you're currently avoiding including previously-included packages, there might be a better time for swift-testing to be included.

My apologies, I think there is a misunderstanding here. When I [talked here](https://forums.swift.org/t/nominations-for-the-packages-community-showcase-on-swift-org/68168/25) about excluding packages for already being nominated and featured, that thread is specifically about the Community Showcase, which is a new set of community-nominated packages every month, and so should not contain duplicates so that everyone can get their turn at being featured.

The other lists on the [Packages page](https://www.swift.org/packages/) are not driven through a voting process at all. They are driven by the Swift Package Index ranking for a keyword or search term. Slowly over time, the packages in these lists do change (as you have seen by swift-testing now rising up the ranks), but generally these show the same or similar packages from month to month as those rankings don't change quickly.

The same data file drives both types of lists, and I think that's where the confusion is coming from. My apologies, we should make it clearer what is happening.

> See [this post](https://forums.swift.org/t/package-swift-testing-is-using-swift-tools-version-5-10-0/69821) for more "official" info about swift-testing's situation.

This is very useful context, thank you. I had not seen that thread. I see we are now not even able to parse the package manifest as we are stuck showing v0.3.0.

In this case, I agree we should remove the swift-testing package out of the package index until we support the version of Swift that it works with. If it was submitted for inclusion into the index today, we would not be able to add it, so I think we should remove it. Tracking this at https://github.com/SwiftPackageIndex/PackageList/pull/6328 

Once the package has been removed, I'll regenerate this file and it will go away.

Thanks so much for helping with this @MahdiBM

/cc @finestructure

daveverwer: swift-testing is now removed from SPI and I have updated this PR to remove it from here, too.

Thanks again, everyone.

_data/packages/packages.yml
daveverwer: This is now fixed.

daveverwer:
@swift-ci please test

daveverwer:
I think this addresses all raised issues and is ready for merging. Just awaiting an approval.

0xTim (approved):

PR #532, opened 2024-02-12
Add top section layout change

federicobucchi:
### Motivation:

Now that the homepage has a wider template we can move the Swift callout (and have it bigger) and the code box in the same row to reduce scrolling and have more content visible (without scrolling)

### Modifications:

Changed the CSS in:
```
assets/stylesheets/_screen.scss
assets/stylesheets/_mobile.scss
assets/stylesheets/pages/_landing.scss
```

Changed the structure of the `callout` section in homepage in:
```
index.md
```

### Result:

Screenshot 2024-02-12 at 1 52 24 PM

Screenshot 2024-02-12 at 1 52 37 PM

Screenshot 2024-02-12 at 2 20 02 PM

### DIFF
--- a/assets/stylesheets/_screen.scss
+++ b/assets/stylesheets/_screen.scss
@@ -385,6 +385,9 @@ main {
     display: grid;
     gap: 1rem;
     grid-template-columns: repeat(2, 1fr);
+    align-items: center;
+    justify-content: center;
+    margin-bottom: 2rem;
 
     @media only screen and (max-width: 1280px) {
       display: block;
--- a/assets/stylesheets/_screen.scss
+++ b/assets/stylesheets/_screen.scss
@@ -848,6 +848,7 @@ code {
 }
 pre code {
   white-space: inherit;
+  font-size: 1.2em;
 }
 
 pre code .graphic {
--- a/assets/stylesheets/_screen.scss
+++ b/assets/stylesheets/_screen.scss
@@ -105,10 +105,6 @@ h6 {
   color: var(--color-header-text);
 }
 
-div.highlighter-rouge {
-  margin-left: 13px;
-}
-
 pre {
   font-size: 14px;
   line-height: 1.6em;
@@ -382,13 +378,25 @@ cite {
 }
 
 main {
-  padding: 4rem 30px 0;
   margin: auto;
+  padding: 4rem 0 0;
+
+  .callout {
+    display: flex;
+
+    @media only screen and (max-width: 1280px) {
+      display: block;
+    }
+  }
 
   .page-wide {
     max-width: var(--content-mobile-breakpoint);
     padding: 0;
     margin: auto;
+
+    @media only screen and (max-width: 1280px) {
+      padding: 30px 30px 0;
+    }
   }
 
   .post, .page {
@@ -396,9 +404,17 @@ main {
     margin: 0 auto;
     padding: 0 30px;
 
-    @media only screen and (max-width: 767px) {
-      padding: 0;
+    @media only screen and (max-width: 1280px) {
+      padding-top: 30px;
     }
+
+    @media only screen and (max-width: 767px) {
+      padding: 30px 30px 0;
+    }
+  }
+
+  @media only screen and (max-width: 1280px) {
+    padding: 0;
   }
 }
 
--- a/assets/stylesheets/_screen.scss
+++ b/assets/stylesheets/_screen.scss
@@ -382,7 +382,9 @@ main {
   padding: 4rem 0 0;
 
   .callout {
-    display: flex;
+    display: grid;
+    gap: 1rem;
+    grid-template-columns: repeat(2, 1fr);
 
     @media only screen and (max-width: 1280px) {
       display: block;
@@ -848,7 +850,6 @@ code {
 }
 pre code {
   white-space: inherit;
-  font-size: 1.2em;
 }
 
 pre code .graphic {
--- a/assets/stylesheets/pages/_landing.scss
+++ b/assets/stylesheets/pages/_landing.scss
@@ -1,5 +1,5 @@
 .preamble {
-  font-size: 38px;
+  font-size: 40px;
   line-height: 1.4;
 
   strong {
--- a/assets/stylesheets/pages/_landing.scss
+++ b/assets/stylesheets/pages/_landing.scss
@@ -1,10 +1,12 @@
 .preamble {
-  font-size: 28px;
+  font-size: 38px;
   line-height: 1.4;
   margin-bottom: 30px;
+  width: 46%;
 
   strong {
     font-weight: 600;
+    color: var(--color-figure-orange);
   }
 
   span {
@@ -12,7 +14,7 @@
     padding-top: .5rem;
   }
 
-  @media only screen and (max-width: 767px) {
+  @media only screen and (max-width: 1280px) {
     width: 100%;
     font-size: 24px;
   }
@@ -20,20 +22,29 @@
 
 .featured-snippet {
   display: none;
+  min-height: 18rem;
+  width: 52%;
+  margin-left: 2%;
 
   &.visible {
     display: block;
   }
 
-  &.highlighter-rouge {
-    margin-left: 0px;
+  .highlight {
+    height: 100%;
   }
 
   pre {
     border: 1px solid var(--color-fill-tertiary);
     border-radius: 4px;
-    padding: 1.25rem 1.5rem;
+    padding: 1rem;
     margin: 0px;
+  }
+
+  @media only screen and (max-width: 1280px) {
+    margin-left: 0;
+    width: 100%;
+    min-height: auto;
   }
 }
 
--- a/assets/stylesheets/pages/_landing.scss
+++ b/assets/stylesheets/pages/_landing.scss
@@ -1,6 +1,7 @@
 .preamble {
   font-size: 40px;
   line-height: 1.4;
+  margin-bottom: 0;
 
   strong {
     font-weight: 600;
@@ -8,12 +9,18 @@
 
   span {
     display: block;
-    padding-top: .5rem;
+    padding-top: 2rem;
   }
 
   @media only screen and (max-width: 1280px) {
     width: 100%;
     font-size: 24px;
+    margin-bottom: 1.6rem;
+
+    span {
+      display: block;
+      padding-top: .5rem;
+    }
   }
 }
 
--- a/assets/stylesheets/pages/_landing.scss
+++ b/assets/stylesheets/pages/_landing.scss
@@ -1,7 +1,7 @@
 .preamble {
   font-size: 38px;
   line-height: 1.4;
-  width: 42%;
+  width: 50%;
 
   strong {
     font-weight: 600;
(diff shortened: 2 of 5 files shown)


federicobucchi:
@swift-ci please test

assets/stylesheets/pages/_landing.scss
```diff
@@ -1,40 +1,51 @@
 .preamble {
-  font-size: 28px;
+  font-size: 38px;
   line-height: 1.4;
   margin-bottom: 30px;
+  width: 46%;
 
   strong {
     font-weight: 600;
+    color: var(--color-figure-orange);
```
shahmishal, line 9: This is back, is it expected?

federicobucchi: Yes, this is correct to be part of this PR for "Top section layout change"

kaishin:
Some initial thoughts:

- Overall I agree that the top section layout needs to be revisited, especially with the recent bump in container width making the introduction line a bit lengthy.
- In the currently proposed solution in the PR, I think the visual balance between the introduction text and the code block is off: either the intro text needs to be made shorter, or the code samples need to be more concise with larger text, or both. As it stands now, the intro text completely eclipses the code samples in visual weight, especially with the orange tint.
  - One potential solution would be to use more new lines in the code samples and bump the text size by 1.5 so as to avoid having a lot of empty space at the bottom of the code block. Not sure this is going to work for all code samples though.
- The use of the brand color is something we need to look at holistically before we introduce it in the UI. As far as I am aware, it is currently only used in the top navigation active page, and some buttons and tags in the package pages. All usages I found are for interactive elements, so using it for emphasis conflicts with that. We can consider underline or another kind of type treatment perhaps?
- Since this PR is primarily visual, I think it'd help to consider some other solutions before committing to one, as I'd imagine making distinct and successive changes to the top section might become disorienting for visitors.

federicobucchi:
@shahmishal @kaishin I did these changes:

- removed the color change and added the underline style
- made code slightly bigger
- removed a code example (because too short)
- changed a code sample (to make it shorter)

Screenshot 2024-02-18 at 8 51 40 PM

Screenshot 2024-02-18 at 8 51 40 PM

federicobucchi:
@swift-ci please test

assets/stylesheets/pages/_landing.scss
```diff
@@ -1,40 +1,50 @@
...
+  margin-left: 1%;
 
   &.visible {
     display: block;
   }
 
-  &.highlighter-rouge {
-    margin-left: 0px;
+  .highlight {
+    height: 100%;
```
kaishin, line 33: Suggestion: Can we center shorter code snippets vertically to make the awkward space below code less noticeable? It can be done using the following properties:
```css
display: flex;
justify-content: space-evenly; 
flex-direction: column;
```
screenshot-2024-02-20-2

The improvement is subjective so please feel free to disagree.

assets/stylesheets/pages/_landing.scss
```diff
@@ -1,40 +1,50 @@
...
+  @media only screen and (max-width: 1280px) {
     width: 100%;
     font-size: 24px;
   }
 }
 
 .featured-snippet {
   display: none;
+  min-height: 18rem;
+  width: 56%;
```
kaishin, line 25: Change request: Let's make this 50%.

Rationale:

I assume that it's set to 56% so that all code snippets would fit horizontally, but I have some reservations:
1. It only works for the default font size.
2. There is only one snippet which doesn't fit at 50-50 width.
3. The ASCII art snippet has a ton of white space to the right as a result.
4. The visual balance/symmetry is thrown off.

As such having a single code snippet slightly cut-off is a good trade-off, especially since the container is scrollable.

assets/stylesheets/pages/_landing.scss
```diff
@@ -1,40 +1,50 @@
 .preamble {
-  font-size: 28px;
+  font-size: 38px;
   line-height: 1.4;
-  margin-bottom: 30px;
+  width: 42%;
```
kaishin, line 4:
```suggestion
  width: 50%;
```

See rationale on the `.featured-snippet` comment.

kaishin (changes requested):
The recent changes are looking great! A couple more layout-related aspects we can address before merging.

assets/stylesheets/pages/_landing.scss
kaishin: As a reference, here is how it looks like with a 50-50 split:
screenshot-2024-02-20

federicobucchi:
@kaishin @cthielen

I applied the suggested changes and I changes ScientificName to Binomial:

> A plant name or binomial is made up of two names: a genus name and a (usually) descriptive specific epithet (species name), both commonly of Latin or Greek origin.

Result:

Screenshot 2024-02-20 at 12 37 06 PM

assets/stylesheets/_screen.scss
```diff
@@ -832,6 +848,7 @@ code {
 }
 pre code {
   white-space: inherit;
+  font-size: 1.2em;
```
alexandersandberg, line 851: This change is now applied site-wide, and not just for the landing page, which we do not want.

Also, this change makes the code snippets look quite... messy? I think it's due to the font size/line height ratio being off. We should also increase the line height if we want to increase the font size.

assets/stylesheets/pages/_landing.scss
```diff
@@ -1,40 +1,52 @@
 .preamble {
-  font-size: 28px;
+  font-size: 38px;
   line-height: 1.4;
-  margin-bottom: 30px;
+  width: 50%;
 
   strong {
     font-weight: 600;
+    text-decoration: underline;
```
alexandersandberg, line 8: I don't think we should underline the words in the preamble. It just makes the page look way more busy than it has to.

assets/stylesheets/_screen.scss
```diff
@@ -382,24 +378,44 @@ cite {
 }
 
 main {
-  padding: 4rem 30px 0;
   margin: auto;
+  padding: 4rem 0 0;
+
+  .callout {
+    display: flex;
```
alexandersandberg, line 385: It would be nice to add a gap here to (1) ensure that the text does not collide with the code snippet and (2) align the code snippet edge with the buttons below.

CleanShot 2024-02-21 at 09 46 38


```suggestion
    display: flex;
    gap: 1rem;
```

assets/stylesheets/_screen.scss
alexandersandberg: Also, I don't think we should increase the font size of the code snippet on mobile, only on desktop—if at all.

assets/stylesheets/pages/_landing.scss
federicobucchi: @alexandersandberg @kaishin - I removed the `underline`

federicobucchi:
@swift-ci please test

dempseyatgithub:
I think using the heavy typeface style already provides enough emphasis for those words.

Adding the underline implies that they might be links, which they aren't, and to me it seems like emphasis overkill.

Two other minor pieces of feedback:

The code area border is flush right with the 'Explore Packages' button, but it hangs over the left edge of the 'Read the docs' button. It seems like the left edge of the code area should be flush with the left edge of the 'Read the docs' button.

If it would be possible to make the title text width just a little smaller, so that "joy" bumps down to the next line, I think it would read better. If that's too much of an issue, I understand. I also understand that we don't have control over how the title will break as the user resizes the window smaller, but we can adjust how it will look at full width.

Thank you - and I think overall this looks very good!

federicobucchi:
@dempseyatgithub I didn't update the screens (+ made the text slightly bigger to have "joy" in a new line), take a look at this:

Screenshot 2024-02-21 at 11 57 52 AM

federicobucchi:
@swift-ci please test

federicobucchi:
@swift-ci please test

alexandersandberg:
At least on my laptop, I feel like this top section is now very big, and takes up a bit too much space.

CleanShot 2024-02-22 at 08 32 58

How do we feel about reducing the code snippet size to its original size, shrinking the preamble text to ~36px, adding an extra margin to the right of the preamble `h1` to make the lines break better (this should not be done by increasing the font size), and increasing the margin between the two preamble sentences?

Something like this:
CleanShot 2024-02-22 at 08 32 41

kaishin (approved):
@federicobucchi Thank you for addressing all the feedback! 

As far as I am concerned this is good to merge. I think @alexandersandberg's suggestion is a net improvement but can be a subsequent PR since it's a rather minor visual change. It's also fine if it's done as part of this PR, up to you @federicobucchi.

I deployed a preview to give this a quick test [here](https://swift-org-website-kaishin.vercel.app).

alexandersandberg:
> I think @alexandersandberg's suggestion is a net improvement but can be a subsequent PR since it's a rather minor visual change. It's also fine if it's done as part of this PR, up to you @federicobucchi.

I think it would still be good to figure it out in this PR (whether we want to apply these suggestions or not), to keep the number of changes to the landing page down, even if minor. 🙂

federicobucchi:
@alexandersandberg @kaishin what about have everything center vertically?  I personally like the vertical centered option and a little breathing within content look good:

Screenshot 2024-02-22 at 10 55 11 AM

Screenshot 2024-02-22 at 10 55 20 AM

Screenshot 2024-02-22 at 10 55 28 AM

kaishin:
@federicobucchi Centering vertically indeed seems like a good solution until we revisit the rotating code blocks. 👍

federicobucchi:
@swift-ci please test

assets/stylesheets/_screen.scss
```diff
@@ -105,10 +105,6 @@ h6 {
   color: var(--color-header-text);
 }
 
-div.highlighter-rouge {
-  margin-left: 13px;
-}
-
```
alexandersandberg, line 111: This left margin is now removed from all code blocks on the site. I don't know if this was intentional.

alexandersandberg (approved):
Looks great, well done! Just one small comment.

assets/stylesheets/_screen.scss
federicobucchi: Yes I did it on purpose.  For the homepage it was adding the margin left that wasn't needed anymore (since not there is a `gap` in between the preamble and the snippet).

For the other pages, it was adding an unnecessary (in my opinion) margin left:

Before:
Screenshot 2024-02-23 at 10 08 53 AM

After:
Screenshot 2024-02-23 at 10 09 10 AM

federicobucchi:
@swift-ci please test

assets/stylesheets/_screen.scss
alexandersandberg: In general, it's better to open a separate PR for changes that are unrelated to the initial scope of the PR, but I think this is fine for now.

I want to propose some changes to this code block design anyway, and will soon open a PR for that. 🙂

federicobucchi: Time I will keep my change, since you did the same change and I need it for my PR: 

https://github.com/apple/swift-org-website/pull/565/files#diff-7d19977d88952efdf1df98667e71041ad606d8831defe8fbd11df7356b2a70ffL108

I don't think at this point make sense for me to remove the margin only for desktop and having you removing the line I added for desktop in this PR, in the PR you made after.

PR #536, opened 2024-02-13
Add NIO on Windows GSoC proposal

finestructure:
cc @compnerd

### DIFF
--- a/gsoc2024/index.md
+++ b/gsoc2024/index.md
@@ -242,6 +242,34 @@ This project is aimed at simplifying the alignement with an ever-evolving SAM te
 
 - [Sebastien Stormacq](mailto:stormacq@amazon.com) | ([GitHub](https://github.com/sebsto)
 
+### Swift NIO on Windows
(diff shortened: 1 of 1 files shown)


FranzBusch (commented):
I think this is a great project but we need to figure out the CI situation for Windows in the NIO repos first cc @compnerd @shahmishal

ktoso:
That's a very good project, we'd need to ensure we have CI so the work isn't wasted and can actually be verified and accepted upstream.

Sounds like we need to chat about Windows CI again if we are to accept such GSoC idea // cc @shahmishal 

I'll see if I can set up a thread where we can talk about this

ktoso:
I discussed the CI situation and with the NIO team, the team is not comfortable mentoring such project without CI so we can't really accept this idea.

We'd love to run it but it's not responsible to do so without CI, maybe next year.

finestructure:
No worries and thanks for following up, Konrad!

PR #554, opened 2024-02-20
Add Swift 6.0 to evolutions list filter

alexandersandberg:
Add version 6.0 to the evolution list's implemented filter.

### Motivation:

As we have some proposals marked as implemented in Swift 6.0, we should add this version to the evolutions filter as well.

### Modifications:

- Adds `6.0` to the filter

### Result:

You can filter the evolutions list by "implemented in Swift 6.0" (currently 8 results).

CleanShot 2024-02-20 at 20 31 56

### DIFF
--- a/assets/javascripts/swift-evolution.js
+++ b/assets/javascripts/swift-evolution.js
@@ -37,6 +37,7 @@ var languageVersions = [
   '5.9',
   '5.9.2',
   '5.10',
+  '6.0',
   'Next'
 ]
 


alexandersandberg:
@swift-ci please test

dempseyatgithub:
The language group is deciding (probably in a day or so) whether we want to move to using 'X.0' for major releases or stick with 'X'.

I've got PRs lined up to make the change in both places depending on which they decide.

In the interim, it looks very strange and inconsistent in the filter UI to have some major versions as '3', '4', '5' and then Swift 6 listed as '6.0'.

dempseyatgithub:
@alexandersandberg I have a PR up for swift-evolution that implements the Language Steering Group decision. https://github.com/apple/swift-evolution/pull/2335

I also have put up https://github.com/apple/swift-org-website/pull/557 here which updates the past versions to match the new style and adds 6.0.

Sorry for the thrash on this.

alexandersandberg:
IMO I still see value in merging this PR now instead of having to wait for the other one which is blocked.

dempseyatgithub (approved):
Yes, I was hoping the PR to swift-evolution would be merged quickly to avoid the inconsistency in the major number versioning in the filter interface.

But it's probably better to have the functionality to filter for the next release even with the version number inconsistency in the interim.

PR #557, opened 2024-02-21
Update past major versions to new '.0' style

dempseyatgithub:
> The LSG decided that we'd like to use 6.0 going forward to avoid confusion with Swift language modes. Updating the existing proposals is the right thing to do. 

Based on the decision of the Language Steering Group, I've put up a PR on swift-evolution to update the implementation versions of past proposals to match using '.0' with major releases. https://github.com/apple/swift-evolution/pull/2335

As soon as that PR lands, this PR will need to be merged so that the version strings match up.


dempseyatgithub:
@swift-ci test

dempseyatgithub:
@swift-ci test

alexandersandberg (approved):
I guess we need to wait with merging this until the evolutions source file has been updated? I'm not sure when that happens exactly.

dempseyatgithub:
Yes, I'll keep an eye out for when the JSON file updates and merge this.  Probably within about an hour or so.

PR #561, opened 2024-02-22
Add archives for the Packages page Community Showcase category

daveverwer:
### Motivation:

As recently discussed with the SWWG, this PR introduces an archive of Community Showcase pages from previous months, since the launch last November.

### Modifications:

- [x] Added a new "Archive" section at the bottom of the "Community Showcase" package list.
- [x] During site generation, generate archive pages for previous month's data based on `_data/packages/history.yml`

### Result:

The Community Showcase page now shows an "Archive" section under the current month's packages:

Screenshot 2024-02-22 at 13 27 43@2x

Each of the archive pages looks like this:

Screenshot 2024-02-22 at 13 31 33@2x


dempseyatgithub (commented):
I just took a brief look at this and was thinking it may make sense to have a single file that contains all of the community package sets including the current set. (Possibly renaming `history.yml` to `community_showcase.yml` or `community.yml`)

The main Community Showcase page would always display the first entry in this list and the archives would always display all but the first entry in the list.

This would make the monthly edit less tedious and error-prone.

It would always be a single additive diff for the new month which simultaneously updates the current Community Showcase and ensures the previous month shows up in the archive.

I realize there would need to be some special-casing on the Community Showcase page as to the source of the package list (using the slug / description from `packages.yml` but the actual list of packages from `history.yml`), but I think that will reduce possible errors (forgetting to add to the archive, miscopying something from `packages.yml` to `history.yml`, etc.)

This is a similar approach to how the most recent release of Swift is based on the last entry of the `swift_releases.yml` file.

daveverwer:
> First - overall this looks great!

Thanks!

> I just took a brief look at this and was thinking it may make sense to have a single file that contains all of the community package sets including the current set.

Ah, there’s probably something about this change that is non-obvious. No human will ever edit these files so I designed them to be easy for implementation here rather than for editing.

These files get generated with a [command line tool](https://github.com/SwiftPackageIndex/PackageListTool) that works from [this source file](https://github.com/SwiftPackageIndex/PackageListTool/blob/main/source.yml), which is much more human editable, and also includes both sets of data in one file. Moving to a new month is a case of moving the current showcase `packages` lines down into the history and adding the month. I kept simplicity in updates very high in my priorities while working on this, as it’s me doing the updating every month.

We did talk about this tool when the original Packages page went through PR, but that was a while ago now and there’s no visibility of the changes I make there every month as they just come through as a PR here already processed.

> (Possibly renaming `history.yml` to `community_showcase.yml` or `community.yml`)

Good idea. Renamed to `showcase-history.yml`.

> The main Community Showcase page would always display the first entry in this list and the archives would always display all but the first entry in the list.

It would be important not to include all of the archives on one page as that page would quickly get too big. I think this method of storing the data might make generating those individual files quite a bit more complex to implement in a static site generator.

> It would always be a single additive diff for the new month which simultaneously updates the current Community Showcase and ensures the previous month shows up in the archive.

You’re correct that the monthly package update diff isn’t very readable, and if that’s a problem then we can deal with that, maybe by making three files. One for the package lists generated by searches, one for the current community showcase, and one for history. We’d still get one difficult diff in the community showcase file, but it would be clearer what’s going on.

That’s a bigger piece of work though, so I’d probably want to tackle that as a separate PR.

dempseyatgithub:
I think the bold "This list is an archive of the..." sentence should be a separate paragraph. It's definitely a separate thought/topic than the rest of the paragraph, and it would better serve as a small descriptive heading for the list.

daveverwer:
> I think the bold "This list is an archive of the..." sentence should be a separate paragraph. It's definitely a separate thought/topic than the rest of the paragraph, and it would better serve as a small descriptive heading for the list.

I agree with this, and even [started to play with a subtitle change to the site](https://github.com/apple/swift-org-website/pull/561/commits/a57956d068c572ad40a98c1b33ad4c96bdfa8383#diff-e4765ee3bb524bb685a32a3c33d84fadead52dc35c7ccea130bcc24b4b2c4302) when developing this, but I decided it was too much to add in one PR. If we need subtitles, we should do that as a separate PR.

For now, I think the best we can do is this, although I have now also added the word Archive. We're getting really tight for space though, and this is with the new wide design!.

Screenshot 2024-02-27 at 17 14 56@2x

dempseyatgithub:
> I agree with this, and even started to play with a subtitle change to the site when developing this, but I decided it was too much to add in one PR. If we need subtitles, we should do that as a separate PR.

Yes, it shouldn't hold up this PR.

> For now, I think the best we can do is this, although I have now also added the word Archive. We're getting really tight for space though, and this is with the new wide design!

 Maybe **Community Showcase: January 2024** would work for the title, with a colon instead of the 'from'.
 
 That would be shorter than the original, still get across the idea that it is in the past. And then defer changes to another PR that adds subtitles.

daveverwer:
@swift-ci please test

shahmishal (approved):

dempseyatgithub (approved):

daveverwer:
@swift-ci test

PR #581, opened 2024-03-06
Updated the Platform Support file.

AbhishekSharma55:
Added {% include_relative %} in platformsupport.md to show updated PlatformSupport content.

Updated PlatformSuppport file so that it shows Latest Content

### Motivation:

As https://swift.org/about/#platform-support is outdated

### Modifications:

Updated the file by removing the previous content and adding the Latest file using Liquid Tag.

### Result:

Now , https://swift.org/about/#platform-support will show the correct Platform Support details.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -25,3 +25,6 @@ xcuserdata
 
 # Prettier
 .prettierrc
+branch_structure.json
+temp_auto_push.bat
+temp_interactive_push.bat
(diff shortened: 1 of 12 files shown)


alexandersandberg (commented):
I'm not sure we want to simply mirror https://www.swift.org/platform-support/ to https://www.swift.org/about/#platform-support. They serve different purposes; on /about you can find a quick overview of where the Swift language is supported, and on /platform-support a deeper dive into the specifics.

We may need to update the content on /about still, but I don't think this proposed solution is how we want to go about it.

AbhishekSharma55:
Hi @alexandersandberg ,

Thank you for your feedback on the proposed update to https://www.swift.org/about/#platform-support. I understand your concern regarding the distinction between the purpose of https://www.swift.org/about/#platform-support and https://www.swift.org/platform-support/.

To ensure we address the issue effectively, could you please provide more clarity on what specific changes or updates you believe would be more suitable for /about?

PR #584, opened 2024-03-07
Package page updates for March

daveverwer:
This update contains the new data file with new packages for the [Packages page](https://www.swift.org/packages/). It also moves last month's packages into the archive.

Packages in the Community Showcase were nominated in [this thread on the Swift forums](https://forums.swift.org/t/nominations-for-the-packages-community-showcase-on-swift-org/68168) and voted on by the [SWWG](https://www.swift.org/website-workgroup/).

All other packages are updated based on search ranking on the [Swift Package Index](https://swiftpackageindex.com/).

### DIFF
--- a/_data/packages/packages.yml
+++ b/_data/packages/packages.yml
@@ -6,14 +6,22 @@ categories:
       nominate packages for consideration in this section."
     description: This showcase celebrates new and interesting packages discussed by
       the Swift community. Package selection in this section is an editorial process
-      organised via [this thread in the Swift Forums](https://forums.swift.org/t/nominations-for-the-packages-community-showcase-on-swift-org/68168)
+      organised via [this thread in the Swift Forums](https://forums.swift.org/t/68168)
       and curated by the [Swift Website Workgroup](https://www.swift.org/website-workgroup/).
     packages:
-      - name: VersionedCodable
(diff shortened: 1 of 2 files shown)


daveverwer:
> I think you meant 'March' instead of 'February' in your initial PR description.

Thank you. Fixed (and using wording so I won't make this mistake next month!) 

> It looks like this PR has changes to other package categories in addition to the Community Showcase. Was that intentional? Is there a process around choosing which packages appear in the other categories? (That shouldn't prevent merging, I just found changes in the other categories to be unexpected.)

Yes, the other package lists are based on search results and rankings from SPI, so this data is always reasonably current in terms of the top ranked packages and compatibility etc… I did mention this in the PR, but I separated it into a separate paragraph now so it's a bit more obvious.

dempseyatgithub:
> Yes, the other package lists are based on search results and rankings from SPI, so this data is always reasonably current in terms of the top ranked packages and compatibility etc… I did mention this in the PR, but I separated it into a separate paragraph now so it's a bit more obvious.

Thanks for the clarification Dave. I had read your originally PR comment, but may have skimmed over and missed that. The PR comment is now very clear (and reusable!)

daveverwer:
Thanks for the review @dempseyatgithub!

I'll merge this in a few hours unless there are any other comments.

PR #585, opened 2024-03-08
Fix typo in Pack Iteration post

simanerush:
- In the initial version of the post, fix that the types of the tuples being compared did not match by making them the same type. 
- Also, fix the call site explanation by explicitly noting that one parameter pack `each Element` represents the types of the two tuples.


simanerush:
@swift-ci test

DougGregor (approved):

cthielen (approved):

PR #586, opened 2024-03-08
Pack Iteration post: Change `Comparable` to `Equatable` in the `==` operator signature

simanerush:
Fix another typo: use `Equatable` in the `==` signatures in the first section. I had a `<` operator there before so I forgot to update the protocol when I changed the operator to `==`


simanerush:
@swift-ci test

cthielen (approved):

PR #598, opened 2024-03-14
Consolidate constant urls at top of script

dempseyatgithub:
Small refactoring PR to consolidate URL constants at the top of the script and make them `const` instead of `var`.

Also add a constant for the URL for the evolution metadata JSON file. This also makes it easier for testing the new JSON file URL during development.

### DIFF
--- a/assets/javascripts/swift-evolution.js
+++ b/assets/javascripts/swift-evolution.js
@@ -9,10 +9,10 @@
 // ===---------------------------------------------------------------------===//
 'use strict'
 
-let EVOLUTION_METADATA_URL = 'https://download.swift.org/swift-evolution/proposals.json'
-let GITHUB_BASE_URL = 'https://github.com/'
-let REPO_PROPOSALS_BASE_URL = GITHUB_BASE_URL + 'apple/swift-evolution/blob/main/proposals'
-let UFF_INFO_URL = '/blog/using-upcoming-feature-flags/'
+const EVOLUTION_METADATA_URL = 'https://download.swift.org/swift-evolution/proposals.json'
+const GITHUB_BASE_URL = 'https://github.com/'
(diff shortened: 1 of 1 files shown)


assets/javascripts/swift-evolution.js
```diff
@@ -9,6 +9,11 @@
 // ===---------------------------------------------------------------------===//
 'use strict'
 
+let EVOLUTION_METADATA_URL = 'https://download.swift.org/swift-evolution/proposals.json'
+let GITHUB_BASE_URL = 'https://github.com/'
+let REPO_PROPOSALS_BASE_URL = GITHUB_BASE_URL + 'apple/swift-evolution/blob/main/proposals'
+let UFF_INFO_URL = '/blog/using-upcoming-feature-flags/'
```
alexandersandberg, line 15: Did you mean `const` here?

```suggestion
const EVOLUTION_METADATA_URL = 'https://download.swift.org/swift-evolution/proposals.json'
const GITHUB_BASE_URL = 'https://github.com/'
const REPO_PROPOSALS_BASE_URL = GITHUB_BASE_URL + 'apple/swift-evolution/blob/main/proposals'
const UFF_INFO_URL = '/blog/using-upcoming-feature-flags/'
```

dempseyatgithub:
@swift-ci test

assets/javascripts/swift-evolution.js
dempseyatgithub: Yes, thanks. I had to redo my original change locally and forgot to make everything const.

Thank you!

federicobucchi (approved):
Ok, next, I really think we need to just do it for `const` and `let` removing all `var`.

dempseyatgithub:
@swift-ci test

Issue #577, opened 2024-03-04
Download menu item / button looks extra big in mobile menu


dempseyatgithub:
The Download menu item looks extra big in the mobile menu.

It ideally would be less prominent.

![image](https://github.com/apple/swift-org-website/assets/470139/6771bd68-c5e7-43e2-b798-c1c72a3409eb)

PR #600, opened 2024-03-18
<577> Fix CTA button in mobile nav

federicobucchi:
### Motivation:

Proposed fix for: https://github.com/apple/swift-org-website/issues/577

### Modifications:

Fix .cta class for mobile navigation style

### Result:

Screenshot 2024-03-17 at 10 09 32 PM

### DIFF
--- a/assets/stylesheets/_screen.scss
+++ b/assets/stylesheets/_screen.scss
@@ -219,7 +219,7 @@ cite {
 
     .active a {
       font-weight: 700;
-      color: var(--color-active-nav-item); 
+      color: var(--color-active-nav-item);
     }
 
     &.nav-section {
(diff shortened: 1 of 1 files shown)


federicobucchi:
@swift-ci please test

dempseyatgithub:
I'm fine with this fix thank you! Could you also update `_includes/navigation.html` so the menu item only reads "Downloads" to match what we do in the desktop nav menu?

This also makes me wonder:

Since the title for "Downloads" is now static, can we add it to 'navigation.yml' instead of making it separate?

We could potentially add a conditional in the html template to make the last navigation item use the `cta` class. Or else add an optional 'class' field to each entry in the YAML file that gets conditionally added if present.

That way the full navigation structure is present in the YAML file and the desktop and mobile versions don't get out of sync.

@kaishin @alexandersandberg

federicobucchi:
@dempseyatgithub done

federicobucchi:
@swift-ci please test

dempseyatgithub (approved):

alexandersandberg (approved):
👍

PR #612, opened 2024-03-23
Update evolution script for proposal changes

dempseyatgithub:
An update to allow the SE Dashboard script to continue to work with the current JSON schema and the new JSON schema.

The feedback on the proposed schema reduced the required change to the status field. Now, only a dot needs to be prepended to the new status state value.

This facilitates development and testing during the transition to the updated schema as detailed here; https://forums.swift.org/t/swift-evolution-metadata-proposed-changes/70779

Tested with both the existing JSON file at:
https://download.swift.org/swift-evolution/proposals.json

And a test version of the proposed schema at:
https://download.swift.org/swift-evolution/v1/evolution.json

From the user's standpoint, there is no observable change in the page.

### DIFF
--- a/assets/javascripts/swift-evolution.js
+++ b/assets/javascripts/swift-evolution.js
@@ -158,7 +158,7 @@ function init() {
   var req = new window.XMLHttpRequest()
 
   req.addEventListener('load', function() {
-    let evolutionMetadata = JSON.parse(req.responseText, flattenStatus)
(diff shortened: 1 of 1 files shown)


dempseyatgithub:
@swift-ci please test

dempseyatgithub:
@swift-ci test

0xTim (approved):

Issue #566, opened 2024-02-27
Allow the layout of the Community Showcase package list archives to cope better with multiple years of data


daveverwer:
The current design for the [Community Showcase packages list archive index](https://www.swift.org/packages/showcase.html#archive) is a plain unordered list:

![Screenshot 2024-02-27 at 20 05 34@2x](https://github.com/apple/swift-org-website/assets/5180/8f7622be-7682-46ba-b3d3-a1463941a1be)

## Describe the solution you'd like

There are many potential solutions to making this list fit better in this space when there are more months of data to display.

The SWWG discussed this, but we don't want to be too prescriptive with this issue. Some solutions that were discussed included having the sections expand by year with detail disclosures, or a dropdown with the year and then a better layout for the months in that year. Both these would work, but there are other options, too.

daveverwer:
This is complete so I'll close this issue. Thanks again @federicobucchi!

PR #614, opened 2024-03-25
<566> Add years in community showcase archive

federicobucchi:
### Motivation:

https://github.com/apple/swift-org-website/issues/566

### Modifications:

- Added years in `showcase-history.yml` and include months in each year.
- Loop years instead of months in the view (`_package-list.html`), then loop the months inside each year)
- Add `JS` to make year tabs interactive and show the correct list of months (after you click on an year).
- Add style to handle the new archive view in `_screen.scss`

### Result:

Screenshot 2024-03-24 at 8 22 36 PM

Screenshot 2024-03-24 at 8 22 44 PM

Screenshot 2024-03-24 at 8 22 56 PM

### DIFF
--- a/_data/packages/packages.yml
+++ b/_data/packages/packages.yml
@@ -207,9 +207,21 @@ categories:
         platform_compatibility:
           - Apple
           - Linux
-        platform_compatibility_tooltip: Apple (iOS, macOS, visionOS) and Linux
+        platform_compatibility_tooltip: Apple (iOS, macOS) and Linux
         license: MIT
         url: https://swiftpackageindex.com/orlandos-nl/MongoKitten
+      - name: SwiftyBeaver
+        description: SwiftyBeaver is a flexible, colorful, lightweight logging library
+          for Swift. It supports console, file, and cloud destinations and is ideal for
+          server-side Swift.
+        owner: SwiftyBeaver
+        swift_compatibility: 5.7+
+        platform_compatibility:
+          - Apple
+          - Linux
+        platform_compatibility_tooltip: Apple (iOS, macOS, watchOS, tvOS) and Linux
+        license: MIT
+        url: https://swiftpackageindex.com/SwiftyBeaver/SwiftyBeaver
       - name: fluent
         description: Fluent helps you work with databases, providing a high-level, type-safe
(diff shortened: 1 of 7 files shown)


federicobucchi:
@swift-ci please test

0xTim (approved):

daveverwer (changes requested):
Thanks for tackling this @federicobucchi! It looks great!

I noticed a couple of things. First, when loading the page with the archive index on it, neither tab is selected so no archives show. I wonder if we should pre-select the 2024 item so there are months available for display.

Next, I think we should use the `aria-pressed` attribute to indicate whether each button is pressed rather than CSS classes. We can target this property in the CSS so that we only have one attribute to toggle and the styling should take care of itself. We should have an `aria-pressed` set explicitly on every button, even ones that have not yet been pressed. This will also mean we don’t need to use JavaScript to do an initial selection as I mentioned above, the HTML for that first button can just have `aria-pressed="true”`.

Then, the `_history.yml` is still expecting only a `months` property in the `showcase-history` data structure and so the history pages are not currently working.

Finally. We’ll need to update https://github.com/SwiftPackageIndex/PackageListTool to also echo this new file structure. Both `packages.yml` and `showcase-history.yml` are produced by this tool, and it may also make sense to make the same structural changes to [this part of `source.yml`](https://github.com/SwiftPackageIndex/PackageListTool/blob/5d84dc5bc1eeec6705157dff2d7808356113cbde/source.yml#L75) to keep the generation of `showcase-history.yml` easy.

federicobucchi:
@swift-ci please test

federicobucchi:
@swift-ci please test

daveverwer (approved):
Thanks for all the changes and improvements @federicobucchi! LGTM

daveverwer:
This merged fine but I don't think it deployed for some reason. See https://www.swift.org/packages/showcase.html#archive still has the old layout.

// @shahmishal @federicobucchi

federicobucchi:
I confirm I cannot see the changes in swift.org

shahmishal:
This should be resolved now

federicobucchi:
@shahmishal @daveverwer I see the changes and all works great.  Thank you

PR #618, opened 2024-03-26
Fix all drop-down items highlighted as active

alexandersandberg:
### Motivation:

I noticed that all drop-down items are highlighted as active if the top-level item is active. This PR fixes that.

### Modifications:

- Moved `.active` class to the `.link-container` div

### Result:

|Before|After|
|---|---|
|CleanShot 2024-03-26 at 18 59 33|CleanShot 2024-03-26 at 18 59 56|


alexandersandberg:
@swift-ci please test

kaishin (approved):
Nice catch!

PR #622, opened 2024-04-01
Added a databases category to the Packages page

daveverwer:
As suggested in [this Swift forums thread](https://forums.swift.org/t/packages-page-on-swift-org-relevance-of-a-persistence-or-databases-category/70868) this PR adds a “Database access” package list to the Packages page.

There is more discussion in [the original thread](https://forums.swift.org/t/packages-page-on-swift-org-relevance-of-a-persistence-or-databases-category/70868).

Screenshot 2024-04-01 at 12 48 21@2x

### DIFF
--- a/_data/packages/packages.yml
+++ b/_data/packages/packages.yml
@@ -488,9 +488,84 @@ categories:
           to log details to the console (and optionally a file) with additional information
           such as date, function name, and line number.
         owner: Dave Wood
+  - name: Database access
+    slug: database
+    brief: We all need to store data somewhere! Find everything from local SQLite databases
+      to connections to server-based databases from your Swift on server apps.
+    description:
+      When looking for something as an alternative to [Core Data](https://developer.apple.com/documentation/coredata)
+      or [SwiftData](https://developer.apple.com/xcode/swiftdata/), look no further
+      than this list of popular database packages.
+    more:
+      title: More database packages
+      url: https://swiftpackageindex.com/keywords/database
+    packages:
+      - name: GRDB
+        description: GRDB is a SQLite toolkit for application development. It provides
+          SQL generation, database observation, robust concurrency, migrations, and raw
+          SQL access.
+        owner: "Gwendal Rou\xE9"
+        swift_compatibility: 5.7+
+        platform_compatibility:
+          - Apple
+          - Linux
+        platform_compatibility_tooltip: Apple (macOS) and Linux
+        license: MIT
+        url: https://swiftpackageindex.com/groue/GRDB.swift
+      - name: MongoKitten
+        description: MongoKitten is a fast, pure Swift MongoDB driver built for Server
+          Side Swift. It supports both MongoDB in server and embedded environments and
+          can be used with Vapor or Hummingbird.
+        owner: Orlandos Technologies - OpenSource
+        swift_compatibility: 5.7+
+        platform_compatibility:
+          - Apple
+          - Linux
+        platform_compatibility_tooltip: Apple (iOS, macOS, visionOS) and Linux
+        license: MIT
+        url: https://swiftpackageindex.com/orlandos-nl/MongoKitten
+      - name: Realm
+        description: Realm is a mobile database for iOS, macOS, tvOS, and watchOS. It
+          offers an object-oriented data model, real-time syncing, and encryption. It
+          integrates with SwiftUI and supports various installation methods.
+        owner: Realm
         swift_compatibility: 5.7+
         platform_compatibility:
           - Apple
         platform_compatibility_tooltip: Apple (iOS, macOS, visionOS, watchOS, tvOS)
+        license: Apache 2.0
+        url: https://swiftpackageindex.com/realm/realm-swift
+      - name: fluent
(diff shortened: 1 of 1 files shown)


_data/packages/packages.yml
```diff
@@ -488,9 +488,84 @@ categories:
           to log details to the console (and optionally a file) with additional information
           such as date, function name, and line number.
         owner: Dave Wood
+  - name: Database access
```
0xTim, line 491: Minor nit - WDYT
```suggestion
  - name: Database Access
```

0xTim (approved):
Query around the casing but LGTM

_data/packages/packages.yml
```diff
@@ -488,9 +488,84 @@ categories:
...
+      - name: GRDB
+        description: GRDB is a SQLite toolkit for application development. It provides
+          SQL generation, database observation, robust concurrency, migrations, and raw
+          SQL access.
+        owner: "Gwendal Rou\xE9"
+        swift_compatibility: 5.7+
+        platform_compatibility:
+          - Apple
+          - Linux
+        platform_compatibility_tooltip: Apple (macOS) and Linux
```
kaishin, line 512: This is tangential to the PR since it's also the case on the SPI page, but why is iOS/tvOS not listed for GRDB even though they are supported by the library?

kaishin (commented):
Nice one! 

I am a bit unsure about the "Access" part of the category title as some of these packages aren't necessarily focused on interfacing with databases.

Would something like  "Databases & Persistence" be acceptable or is it a mouthful? Otherwise just "Databases" would cast a wider net in my view.

_data/packages/packages.yml
groue: Good question @kaishin 👍 I think it's because the GRDB repository comes with 2 Xcode projects - this trips up SPI: https://swiftpackageindex.com/groue/GRDB.swift/builds

@daveverwer, do you think you can manually update the `platform_compatibility_tooltip`  with iOS, macOS, tvOS, visionOS, watchOS, and Linux?

daveverwer: Unfortunately manually updating this is not going to be possible. The data files that drive these pages are automatically generated from SPI data and so every time that tool gets run (which happens monthly) it would overwrite any manual changes.

I'd much rather get this fixed on the SPI side, so that it's not only correct here but also on SPI. I vaguely remember this coming up before, did we already go through troubleshooting the builds? If you want to [open a new "Incorrect package data" issue](https://github.com/SwiftPackageIndex/SwiftPackageIndex-Server/issues/new/choose) and we can look at what's causing the failures. We do have some options you can specify in your `.spi.yml` that are able to solve common problems with builds.

dempseyatgithub (approved):
This LGTM!

federicobucchi (approved):

_data/packages/packages.yml
daveverwer: Apologies for not getting back to you on this. @finestructure and I chatted about the compatibility issue and while we do still plan to add explicit support for multiple projects, unfortunately, the change needed touches multiple parts of our application and isn't something we can rush through for this.

There is a [potential workaround](https://github.com/SwiftPackageIndex/SwiftPackageIndex-Server/issues/1890#issuecomment-1191374809) in your original issue. We understand that adopting a workaround like this is not ideal, and of course, we will get to https://github.com/SwiftPackageIndex/SwiftPackageIndex-Server/issues/1915 as soon as possible, but I can't say when as we have not yet had reports of any other packages in need of this feature.

I’m sorry I don’t have better news!

groue: I can't move the project because this would be breaking change for users.

I can't gather all targets in a single Xcode project (in a future major release) because one of the projects is used for advances use cases (custom SQLite builds) and depends on a submodule that not everybody clones. When all targets are gathered in a single project, the missing submodule creates an awful ton of errors and warnings in Xcode - not something I want users to see.

> I can't say when as we have not yet had reports of any other packages in need of this feature.

Well understood, @daveverwer. I can only appeal to your generosity, or the itch created by this wrong meta-data.

If GRDB is well-appreciated, it is also because it goes far in servicing the various needs of its users - hence the two Xcode projects. My priority is and will remain GRDB users.

parispittman:
lgtm!

packages/_package-lists.html
```diff
@@ -8,12 +8,12 @@ <h3>{{ category.name }}</h3>
...
-  <li>
-    <p>
-      There are thousands more packages to browse at the
-      <a href="https://swiftpackageindex.com" target="_blank">Swift Package Index</a>, along with all the metadata you
-      need to make better decisions about your dependencies.
-    </p>
-    <a href="https://swiftpackageindex.com" target="_blank" class="cta-secondary external">Browse more packages</a>
-  </li>
 </ul>
+<p>
```
daveverwer, line 12: This `<p>` needed to either be lifted out of the containing list or the CSS needed modification to allow it to take up a whole line. This seemed the neater option.

groue:
It's online: https://www.swift.org/packages/database.html Thank you! ❤️

PR #635, opened 2024-04-10
Improve page layout with <footer> element

alexandersandberg:
### Motivation:

We're using the `<footer>` element on `article` layout pages for the author bylines. This is styled with some margin and a divider between the content and the authors.

We should use this on `page` layout pages too, to keep the styling consistent.

Using a `<footer>` element here will also improve the semantics of the HTML for the `page` layout.

### Modifications:

- Replaced the `<hr>` in the `page` layout with a `<footer>`

### Result:

|Before|After|
|---|---|
|CleanShot 2024-04-10 at 20 58 12|CleanShot 2024-04-10 at 20 57 51|

Which now looks the same as the blog posts' footer:

### DIFF
--- a/_layouts/page.html
+++ b/_layouts/page.html
@@ -9,10 +9,13 @@ layout: default
       </header>
     {% endif %}
   {% endunless %}
+
   {{ content }}
+
   {% if page.author %}
-    <hr>
-    <p>Contributed by</p>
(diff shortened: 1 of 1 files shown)


alexandersandberg:
@swift-ci please test

0xTim (approved):

PR #646, opened 2024-04-22
Migrate dashboard script to new status values

dempseyatgithub:
This PR migrates the SE Dashboard script to use the new state values for proposal status. 

Essentially values previously had a leading dot (i.e. ".accepted") and now do not (i.e. "accepted")

Details of the schema change is available in the proposal:
https://forums.swift.org/t/swift-evolution-metadata-proposed-changes/70779

This PR does two main things:
1. Moves to using string constants for proposal state values instead of bare strings throughout the script
   - The string constants are defined in a frozen `State` object
  
2. Migrates to the new state values
   - Updates the string contents
   - Removes the reviver function that added a leading dot. This allowed the script to work with both the old and new schema, but is no longer required. 

I have tested the change, including filters, and behavior is unchanged as expected with no errors or warnings logged to the JS console.

### DIFF
--- a/assets/javascripts/swift-evolution.js
+++ b/assets/javascripts/swift-evolution.js
@@ -27,17 +27,17 @@ let upcomingFeatureFlagFilterEnabled = false
 
 /** Proposal state string constants */
 const State = Object.freeze({
-  "awaitingReview": 'awaitingReview',
-  "scheduledForReview": 'scheduledForReview',
-  "activeReview": 'activeReview',
-  "returnedForRevision": 'returnedForRevision',
-  "withdrawn": 'withdrawn',
-  "accepted": 'accepted',
-  "acceptedWithRevisions": 'acceptedWithRevisions',
-  "rejected": 'rejected',
-  "implemented": 'implemented',
-  "previewing": 'previewing',
-  "error": 'error',
+  awaitingReview: 'awaitingReview',
+  scheduledForReview: 'scheduledForReview',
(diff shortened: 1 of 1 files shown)


assets/javascripts/swift-evolution.js
```diff
@@ -25,7 +25,24 @@ let filterSelection = []
...
+  "activeReview": 'activeReview',
+  "returnedForRevision": 'returnedForRevision',
+  "withdrawn": 'withdrawn',
+  "accepted": 'accepted',
+  "acceptedWithRevisions": 'acceptedWithRevisions',
+  "rejected": 'rejected',
+  "implemented": 'implemented',
+  "previewing": 'previewing',
+  "error": 'error',
+})
```
federicobucchi, line 41: This would be correct too:

```
const State = Object.freeze({
  awaitingReview: 'awaitingReview',
  scheduledForReview: 'scheduledForReview',
  activeReview: 'activeReview',
  returnedForRevision: 'returnedForRevision',
  withdrawn: 'withdrawn',
  accepted: 'accepted',
  acceptedWithRevisions: 'acceptedWithRevisions',
  rejected: 'rejected',
  implemented: 'implemented',
  previewing: 'previewing',
  error: 'error',
})
```

federicobucchi (approved):

dempseyatgithub:
@swift-ci test

kaishin (approved):
Much better!

PR #652, opened 2024-05-08
Fix empty 'Written by' headline on blog post contributions page

alexandersandberg:
### Motivation:

Fixes an empty "Written by" headline in the footer on https://www.swift.org/blog-post-contributions/

### Modifications:

- Only show the 'Written by' headline in the footer if the post has one or more authors
- Change the layout for /blog-post-contributions/ to `page`, as it better suits the content

### Result:

- No empty footer on /blog-post-contributions/
- Other posts with authors have not changed
- In case we use the `post` layout for a page that doesn't have an author in the future, this empty headline will no longer be shown

### DIFF
--- a/_layouts/post.html
+++ b/_layouts/post.html
@@ -13,8 +13,10 @@ layout: base
   {{ content }}
 
   <footer>
-    <p>Written by</p>
-    {% include authors.html authors=page.author %}
+    {% if page.author %}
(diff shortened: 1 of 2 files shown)


alexandersandberg:
@swift-ci please test

daveverwer (approved):

PR #669, opened 2024-05-22
[Articles] Add Emacs article.

al45tair:
Add an article showing how to get started with Swift using Emacs.

rdar://127610658

### DIFF
--- a/documentation/articles/zero-to-swift-emacs.md
+++ b/documentation/articles/zero-to-swift-emacs.md
@@ -44,14 +44,14 @@ will install Emacs and set up some shortcuts for you.
 
 From a prompt, enter
 
-```shell
+```console
 $ sudo apt-get install emacs
 ```
 
 or if you are doing this on a server system or in a container where
 you have no plans to use the GUI,
 
-```shell
+```console
 $ sudo apt-get install emacs-nox
 ```
 
@@ -59,7 +59,7 @@ $ sudo apt-get install emacs-nox
 
 From a prompt, enter
 
-```shell
+```console
 $ sudo dnf install emacs
 ```
 
@@ -92,7 +92,7 @@ website.  You can un-tar that in a location of your choice,
 e.g. `/usr/local`.  You may also wish to make a symbolic link to make
 it easier to remember where things are.
 
-```shell
+```console
 $ curl -L -O https://download.swift.org/swift-5.10-release/ubuntu2204/swift-5.10-RELEASE/swift-5.10-RELEASE-ubuntu22.04.tar.gz
 $ cd /usr/local
 $ sudo tar xvzf ~/swift-5.10-RELEASE-ubuntu22.04.tar.gz
--- a/documentation/articles/zero-to-swift-emacs.md
+++ b/documentation/articles/zero-to-swift-emacs.md
@@ -18,6 +18,11 @@ started.  It also is not intended to teach you to use Emacs; if you
 want to learn Emacs, try this resource: [Emacs
 Tour](https://www.gnu.org/software/emacs/tour/).
 
+We assume for the purposes of this guide that you have already
+installed Swift; if not, please [follow the installation instructions
+for your platform on the Swift website](https://www.swift.org/install)
+before continuing.
+
 ## Installing Emacs
 
 There are generic [installation instructions on the GNU Emacs
@@ -65,51 +70,6 @@ $ sudo dnf install emacs
 
 (On older builds you may need to use `yum` rather than `dnf`, but the
 latter is the newer, better option if it's available.)
-
-## Installing Swift
-
-There are [instructions on the Swift
-website](https://swift.org/install).  Again, specific instructions are
-given below for some common platforms.
-
-### macOS
-
-On macOS, it's probably easiest to [install Xcode from the App
-Store](https://itunes.apple.com/app/xcode/id497799835).  It is
-possible to install a packaged toolchain from Swift.org, but if you do
-that then you won't be able to submit applications to the App Store
-(assuming that is something you want to do).
-
-### Microsoft Windows
-
-Use [the Windows installer from the Swift website](https://www.swift.org/download/#releases); that should set up everything you need to get going with Swift.
-
-### Linux
-
-Start by downloading [the tarball for your
-distribution](https://www.swift.org/download/#releases) from the Swift
-website.  You can un-tar that in a location of your choice,
-e.g. `/usr/local`.  You may also wish to make a symbolic link to make
-it easier to remember where things are.
-
-```console
-$ curl -L -O https://download.swift.org/swift-5.10-release/ubuntu2204/swift-5.10-RELEASE/swift-5.10-RELEASE-ubuntu22.04.tar.gz
-$ cd /usr/local
-$ sudo tar xvzf ~/swift-5.10-RELEASE-ubuntu22.04.tar.gz
-$ sudo ln -s swift-5.10-RELEASE-ubuntu22.04 swift
-```
-
-You'll also want to add Swift to your `PATH`.  Exactly how to do this
-depends on which shell you're using, but typically you'll want to edit
-your `.bashrc` or `.zshrc` to add
-
-```shell
-export PATH=/usr/local/swift/usr/bin
-```
-
-or similar.
-
-Before using Swift, you will need to install some dependencies; there is [a set of canned commands for various distributions here](https://www.swift.org/install/linux/#installation-via-tarball).  (Note: the table scrolls left and right.)
 
 ## Configuring Emacs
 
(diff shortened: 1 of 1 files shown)


documentation/articles/zero-to-swift-emacs.md
```diff
@@ -0,0 +1,407 @@
+# From Zero to Swift | Getting started with Swift in Emacs
+
+Emacs is a highly customizable text editor that has its
```
RobinBateman808, line 3: Recommend spelling out Emac on the first mention and link to an educational resource to allow users to learn more about it if they want. Maybe something like, "Editor Macros, also known as [Emacs](https://en.wikipedia.org/wiki/Emacs), is a highly customizable editor...."

documentation/articles/zero-to-swift-emacs.md
```diff
@@ -0,0 +1,407 @@
...
+origins in a macro package for the DEC TECO editor.  A number of
```
RobinBateman808, line 4: Consider spelling out what DEC and TECO mean on the first mention—for example, Digital Equipment Corporation (DEC) Text Editor and Corrector (TECO). Then you can use acronyms throughout the rest of the document.

documentation/articles/zero-to-swift-emacs.md
```diff
@@ -0,0 +1,407 @@
...
+Emacs-like editors and derivatives have existed over the
+years, but this guide will focus on GNU Emacs.
```
RobinBateman808, line 6: Same for GNU ... "...but this guide will focus on [GNU's Not Unix (GNU)](https://en.wikipedia.org/wiki/GNU) Emacs."

documentation/articles/zero-to-swift-emacs.md
```diff
@@ -0,0 +1,407 @@
...
+This guide will walk you through setting up a new install of Emacs for
+Swift development; it does not assume any particular operating system,
+but where relevant may give system-specific hints to help you get
+started.  It also is not intended to teach you to use Emacs; if you
+want to learn Emacs, try this resource: [Emacs
+Tour](https://www.gnu.org/software/emacs/tour/).
+
+## Installing Emacs
+
+There are generic [installation instructions on the GNU Emacs
```
RobinBateman808, line 17: Consider starting this sentence with an action-oriented word-for example, "Follow this guide to walk you through..."

documentation/articles/zero-to-swift-emacs.md
```diff
@@ -0,0 +1,407 @@
...
+website](https://www.gnu.org/software/emacs/download.html).  Specific
+instructions are given below for some common platforms.
+
+### macOS
+
+The [Emacs for OSX](https://emacsformacosx.com) website provides
```
RobinBateman808, line 23: Recommend changing the wording to **Emacs for Mac OSX** to match the actual page title for clarity.

documentation/articles/zero-to-swift-emacs.md
```diff
@@ -0,0 +1,407 @@
...
+universal binaries in a standard Mac disk image.  Downloading the
+image from that site, opening it and dragging it to your Applications
```
RobinBateman808, line 25: Consider referencing as `Applications` if it's a UI term. If referring to a placeholder, you can use italics _Applications_.

documentation/articles/zero-to-swift-emacs.md
```diff
@@ -0,0 +1,407 @@
...
+
+On macOS, it's probably easiest to [install Xcode from the App
+Store](https://itunes.apple.com/app/xcode/id497799835).  It is
+possible to install a packaged toolchain from Swift.org, but if you do
+that then you won't be able to submit applications to the App Store
+(assuming that is something you want to do).
+
+### Microsoft Windows
+
+There is [a Windows installer on the Swift
```
RobinBateman808, line 79: Again, recommend starting the sentence with an action-oriented word: "Use the Windows installer on the Swift website..."

documentation/articles/zero-to-swift-emacs.md
```diff
@@ -0,0 +1,407 @@
...
+```shell
+$ export PATH=/usr/local/swift/usr/bin
+```
+
+or similar.
+
+Before using Swift, you will need to install some dependencies; the
+exact list depends on your distribution; some examples are given below.
+
+#### Ubuntu 18.04
```
RobinBateman808, line 111: Maybe consider linking to the currently published [dependencies for Linux](https://www.swift.org/install/linux/#installation-via-tarball) instead of duplicating?

documentation/articles/zero-to-swift-emacs.md
```diff
@@ -0,0 +1,407 @@
...
+         python3-devel \
+         rsync         \
+         sqlite-devel  \
+         unzip         \
+         zip
+```
+
+## Configuring Emacs
+
+Emacs veterans probably remember downloading and installing Lisp
```
RobinBateman808, line 194: Consider defining Lisp as "...installing list processing (Lisp) packages to easily manipulate data strings, but these days..."

documentation/articles/zero-to-swift-emacs.md
al45tair: DEC is possibly worth expanding, even though it no longer exists as an entity, but expanding TECO is I think a bit pointless and doesn't really add any useful information.

documentation/articles/zero-to-swift-emacs.md
al45tair: I don't think we need to expand GNU here, honestly. The audience this article is written for will either already know what it stands for, or actually not care. Also, GNU Emacs is the name of the editor; it isn't just Emacs (since there's Lucid Emacs, XEmacs, Microemacs and probably many others), and it isn't called "GNU's Not UNIX Emacs".

documentation/articles/zero-to-swift-emacs.md
al45tair: Except 99% of users are going to _not_ want to follow that guide but instead use the instructions we've got immediately below. I think we should leave this as-is.

documentation/articles/zero-to-swift-emacs.md
al45tair: Seems reasonable to highlight it somehow. I'll go with `Applications` for now.

documentation/articles/zero-to-swift-emacs.md
```diff
@@ -0,0 +1,407 @@
...
+packages manually, but these days it's best to use a package manager.
+We'll also configure MELPA, which is a popular package repository for
```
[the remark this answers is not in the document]
al45tair: The link is a good idea. Expanding it, not so much IMO (it doesn't really matter what it stands for, and "Milkypostman" lends an unserious air to the entire thing).

documentation/articles/zero-to-swift-emacs.md
RobinBateman808: Ack. Thanks.

0xTim (changes requested):
Overall looks good. The article needs the Jekyll metadata to be able to do a proper review though as I can't see the generated content without it (no HTML page is created)

documentation/articles/zero-to-swift-emacs.md
```diff
@@ -0,0 +1,330 @@
...
+From a prompt, enter
+
+```shell
+$ sudo dnf install emacs
+```
+
+(On older builds you may need to use `yum` rather than `dnf`, but the
+latter is the newer, better option if it's available.)
+
+## Installing Swift
```
0xTim, line 69: I'm in two minds about including this section - it could get out of date pretty quickly, especially with the work being done in #676. I'd probably vote for just linking to that page

documentation/articles/zero-to-swift-emacs.md
```diff
@@ -0,0 +1,330 @@
...
+server](http://ftp.gnu.org/gnu/emacs/windows/).  The simplest thing is
+likely to run the `emacs-<version>-installer.exe` executable, which
+will install Emacs and set up some shortcuts for you.
+
+### Debian-based Linux (including Ubuntu)
+
+From a prompt, enter
+
+```shell
+$ sudo apt-get install emacs
```
0xTim, line 48: Not am massive fan of including the `$` - it makes copy and pasting commands harder and isn't consistent with shell commands across the rest of the site

0xTim (approved):
Overall looks good - some queries around code blocks and we should probably have a bigger discussion around how to link to installing Swift

al45tair:
@swift-ci Please test

al45tair:
@swift-ci Please test

documentation/articles/zero-to-swift-emacs.md
al45tair: I'm happy to rewrite this however anyone wants, and if linking elsewhere is sufficient, that's great.

bnbarham: +1 for just linking out to the install page

al45tair: I've updated to link out to the install page.

al45tair:
@swift-ci Please test

PR #687, opened 2024-06-04
Bundle update to get `rexml` updated after CVE-2024-35176

daveverwer:
Fixes https://github.com/apple/swift-org-website/security/dependabot/2

I don't believe we were at risk of any consequences from this CVE as we don't run anything at runtime, but we've not had a bundle update in a while.


daveverwer:
@swift-ci please test

shahmishal (approved):

PR #676, opened 2024-05-27
Re-design the install page to help guide the user pick the correct installer/tarball

shahmishal:
Users can currently become confused about which toolchain to install for their platform. This change will help guide users in selecting the correct installer or tarball for their platform. Additionally, the page will now automatically detect the platform from which you are visiting.

Screenshot 2024-06-03 at 12 28 43 PM

PR #692, opened 2024-06-06
Improve Linux platform selection copy on install page

alexandersandberg:
### Motivation:

Small proposal for changing the title here to be more concise and consistent, and I don't think we need to mention installing a "specific toolchain" here.

### Modifications:

Updated the copy.

### Result:

|Before|After|
|---|---|
|||


alexandersandberg:
Follow-up to #676. cc: @shahmishal

alexandersandberg:
@swift-ci please test

shahmishal (approved):

PR #694, opened 2024-06-06
Mark prompts as non-selectable.

al45tair:
This means that we can have nicely formatted sequences of shell commands with output without it being unclear, while also making it easy to copy just the command without getting the prompt as well.

Here's what it looks like in action:


al45tair:
@swift-ci Please test

etcwilde (commented):
From what I can tell, `.go` is the output from the command from the `console` lexer. Do we want to make the output unselectable so that folks can just copy-paste the entire text box?

alexandersandberg (approved):
Great idea!

PR #708, opened 2024-06-10
Fix horizontal scroll on mobile on some pages

alexandersandberg:
### Motivation:

On some pages, e.g. https://www.swift.org/documentation/concurrency/, the page scrolls horizontally due to some inline code blocks not breaking. This is not the best experience. Instead, it's better to wrap these code blocks into new lines.

### Modifications:

- Make inline code blocks break and wrap into new lines instead of extending outside the viewport
- Remove inline code blocks from some links on https://www.swift.org/contributor-experience-workgroup/
  - I originally intended to just fix these but then noticed the issue in more places, which is why I applied the site-wide fix instead. However, I decided to still remove these inline code blocks here to keep it consistent with the workgroup contact links in other places, e.g. on https://www.swift.org/website-workgroup/#communication
- Added hyphenation to titles on the smallest viewports (up to 480px) to fix some additional cases of horizontal scrolling
  - I did not want to apply this everywhere as it's sometimes a bit too aggressive (adding a hyphen instead of wrapping the word to a new line), but AFAIK this was the only option to fix this on mobile without design changes (which we should consider, but for another time).

### Result:

|Before|After|
|---|---|
|CleanShot 2024-06-10 at 08 51 21|CleanShot 2024-06-10 at 08 52 20|
|CleanShot 2024-06-10 at 09 04 44|CleanShot 2024-06-10 at 09 05 10|
|CleanShot 2024-06-10 at 09 04 13|CleanShot 2024-06-10 at 09 03 55|

### DIFF
--- a/assets/stylesheets/_mobile.scss
+++ b/assets/stylesheets/_mobile.scss
@@ -218,6 +218,17 @@
   }
 }
 
+@media only screen and (max-width: 480px) {
+  h1,
+  h2,
+  h3,
+  h4,
+  h5,
+  h6 {
+    hyphens: auto;
+  }
+}
+
 .nav-menu-container {
   display: grid;
   grid-template-columns: 1fr;
--- a/assets/stylesheets/_screen.scss
+++ b/assets/stylesheets/_screen.scss
@@ -125,7 +125,7 @@ td {
     padding: 0;
     margin: 0;
     font-size: 16px;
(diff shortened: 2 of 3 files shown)


alexandersandberg:
@swift-ci please test

kaishin (approved):
Love the `hyphens` and `word-break` solution!

PR #709, opened 2024-06-10
Preselect Linux version on /install if only one version available

alexandersandberg:
### Motivation:

Suggestion from #676 by @daveverwer.

### Modifications:

- Add Linux version to the Linux platform tab if only one version is available to select from

Note: The best implementation of this would be to store the platforms/versions in a data file and dynamically (at build-time) decide whether we should directly link to the version or not based on the data.

I did start working on this but decided to go for this simpler, manual implementation for now as I think in general the install page data can be reworked to not only store the platforms/versions in a data file but also _generate the pages_ based on this data.

@shahmishal @federicobucchi let me know what you think about this—whether this quick implementation is still worth it or if we should consider the data file solution right away.

### Result:

On /install, clicking a Linux platform with only one version will directly link to that version instead of requiring an extra click.

### DIFF
--- a/_includes/install/_linux_platforms_tabs.md
+++ b/_includes/install/_linux_platforms_tabs.md
@@ -2,11 +2,11 @@
 
 <div class="interactive-tabs os">
   <div class="tabs">
-    <a href="/install/linux/amazonlinux#versions" aria-pressed="{{ include.amazonlinux }}">Amazon Linux</a>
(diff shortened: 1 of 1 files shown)


alexandersandberg:
@swift-ci please test

0xTim (approved):

shahmishal (approved):

federicobucchi (approved):

PR #720, opened 2024-06-12
Swift.org install api to support installer tools (Swiftly)

shahmishal:
Swift.org install api to support installer tools (Swiftly)

To support Swiftly, I propose static JSON API endpoints from YAML files that will be created during the Jekyll build phase. These endpoints will allow users to access a list of releases at `swift.org/api/v1/install/releases.json`, containing information about tags, platforms, architectures, and docker tags. The dev endpoint, `swift.org/api/v1/install/dev/<branch>/<platform>.json`, will provide a list of development toolchain information. In the future, we should provide REST APIs which can be dynamic and supports query parameters. 

```
├── api
│   └── v1
│       └── install
│           ├── dev
│           │   ├── 6.0
│           │   │   ├── amazonlinux2.json
│           │   │   ├── centos7.json
│           │   │   ├── macos.json
│           │   │   ├── static-sdk.json
│           │   │   ├── ubi9.json
│           │   │   ├── ubuntu2004.json
│           │   │   ├── ubuntu2204.json
│           │   │   └── windows10.json
│           │   └── main
│           │       ├── amazonlinux2.json
│           │       ├── centos7.json
│           │       ├── macos.json
│           │       ├── static-sdk.json
│           │       ├── ubi9.json
│           │       ├── ubuntu2004.json
│           │       ├── ubuntu2204.json
│           │       └── windows10.json
│           └── releases.json
```

### DIFF
--- a/_data/builds/swift_releases.yml
+++ b/_data/builds/swift_releases.yml
@@ -1,4 +1,4 @@
-- name: 2.2
+- name: "2.2"
   tag: swift-2.2-RELEASE
   date: 2016-03-21
   xcode: Xcode 7.3
@@ -13,7 +13,7 @@
       platform: Linux
       archs:
         - x86_64
-- name: 2.2.1
+- name: "2.2.1"
   tag: swift-2.2.1-RELEASE
   xcode: Xcode 7.3.1
   xcode_release: true
@@ -28,7 +28,7 @@
       platform: Linux
       archs:
         - x86_64
-- name: 3.0
+- name: "3.0"
   tag: swift-3.0-RELEASE
   xcode: Xcode 8
   xcode_release: true
@@ -42,7 +42,7 @@
       platform: Linux
       archs:
         - x86_64
-- name: 3.0.1
+- name: "3.0.1"
   tag: swift-3.0.1-RELEASE
   xcode: Xcode 8.1
   xcode_release: true
@@ -60,7 +60,7 @@
       platform: Linux
       archs:
         - x86_64
-- name: 3.0.2
+- name: "3.0.2"
   tag: swift-3.0.2-RELEASE
   xcode: Xcode 8.2
   xcode_release: true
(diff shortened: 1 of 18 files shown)


alexandersandberg (commented):
It would be amazing if we could use e.g. https://github.com/avillafiorita/jekyll-datapage_gen to dynamically generate these endpoints based on a single source of truth data file—I guess `_data/builds/swift_releases.yml` would be enough?

This data file could then also be used to generate the pages for /install, instead of manually creating them as well as its content for each new supported platform/version, since this can easily lead to inconsistencies.

shahmishal:
In this pull request, the data are generated from the yml file during the build phase so it would signal source of truth. 

for example, its reading the value from the _data and converting it to json. 
```
{{ site.data.builds.swift-6_0-branch.amazonlinux2-aarch64 | jsonify}}
```

alexandersandberg:
> In this pull request, the data are generated from the yml file during the build phase so it would signal source of truth.
> 
> for example, its reading the value from the _data and converting it to json.
> 
> ```
> {{ site.data.builds.swift-6_0-branch.amazonlinux2-aarch64 | jsonify}}
> ```

I mean to generate the "template" endpoint files for each version as well, so we don't have to create a new `api/v1/install/dev/6.0/{platform/version}.json` JSON file when we add support for more platforms/versions.

I guess now we would also have to duplicate the platform/version JSON files for each new Swift version as well. If they were generated, adding support for a future Swift version would be as simple as only updating the source of truth data file.

shahmishal:
I will look into https://github.com/avillafiorita/jekyll-datapage_gen plugin, I noticed the repo has not been updated for 2 years (it might be fine).

kaishin (approved):
This works well enough for the time being, even though I agree with @alexandersandberg that we can benefit from generating the JSON files dynamically.

0xTim (approved):
Looks like a great step in the right direction! Do appear to be missing the newer platforms though

shahmishal:
>Looks like a great step in the right direction! Do appear to be missing the newer platforms though 
The new platforms are available part of the releases.json only because nightly toolchains currently have an issue with new platforms on main and release/6.0 branches.

shahmishal:
@swift-ci test

PR #727, opened 2024-06-17
Fix footer position

federicobucchi:
### Motivation:

When the content is the page is short, the footer aligns in too high.

### Modifications:

Added style changes in _screen.scss

### Result:

Before

Screenshot 2024-06-17 at 10 36 29 AM

After

Screenshot 2024-06-17 at 10 36 04 AM

### DIFF
--- a/assets/stylesheets/_screen.scss
+++ b/assets/stylesheets/_screen.scss
@@ -8,7 +8,10 @@
 
 html,
 body {
-  height: 100%;
+  display: flex;
+  flex-direction: column;
+  position: relative;
+  min-height: 100vh;
 }
 
 body {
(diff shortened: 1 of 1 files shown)


federicobucchi:
@swift-ci please test

0xTim (approved):

alexandersandberg (approved):
Thanks! Looks good! Tested on mobile as well.

The only thing I noticed is that you will always have a scrolling page on mobile (at least in Safari) due to the `vh` unit, even if the content doesn't expand outside the viewport. However, this is not an issue as (1) we don't have any pages that have this little content and (2) it's still better than having the footer float in the middle of the page. Just wanted to mention it anyway.

https://github.com/swiftlang/swift-org-website/assets/35671299/75ba42bb-a933-45ab-a3f8-2b5dbdab1bcb

PR #741, opened 2024-06-25
Standardize on wide pages

rauhul:
Updates the page templates to use "wide" pages everywhere instead of mix of wide and narrow pages.


rauhul:
@swift-ci test

federicobucchi:
Didn't we decide to keep the pages with a lot of text narrow (for not having the user to have difficulty on reading) and the ones that allow a "multiple columns" template to have the wide template?

rauhul:
Is this something than can be revisited? The different page width look more like a bug than an intentional choice. Is there some prior art of language sites that act like this?

federicobucchi:
We can chat about it, but it would take a longer discussion.  But, just wanted to let you know, that this was decided for the reason above.  Making all the pages wide, does bring real problems for readers.

rauhul:
How does one join that discussion?

federicobucchi:
We can have the conversation in here, or in https://forums.swift.org

For reference: https://github.com/swiftlang/swift-org-website/pull/499#issuecomment-1913094160

rauhul:
I see, I think the points presented there are fair enough, and probably not worth pushing for a better solution right now

federicobucchi (changes requested):
Ok, please feel free to close the PR

PR #749, opened 2024-06-28
Reduce duplicated code for banners

rauhul:
Unifies the banner and cta css styles used on multiple pages.

Moves the element to a common location so the it can be found in the
source easily and be reused across pages.

PR #743, opened 2024-06-25
Fix duplicated grays

rauhul:
Removes figure-gray-secondary-alt in favor of figure-gray-tertiary which already had the exact same color values.

Fixes fill-quaternary to have different colors in dark and light mode.

PR #751, opened 2024-06-28
Modernizes visual appearance of tabs

rauhul:
Updates the .interactive-tabs class to take a more layered appearance.
Hover and selection states are visually distinct and rely on the
existing color palette for layers.

Adds support for "prefers-contrast: more" to use the previous higher
contrast style.

Moves the css for tabs to a common location so it can be easily found
and reused between pages.

Builds on top of #743 and #749


rauhul:
# After <- Before

> Note: I messed up the order of the windows when taking screenshots so the new style is the left side and old style is right.


## Light Default
Screenshot 2024-06-28 at 1 10 29 PM

## Light Increase Contrast
Screenshot 2024-06-28 at 1 09 02 PM

## Dark Default
Screenshot 2024-06-28 at 1 10 20 PM

## Dark Increase Contrast
Screenshot 2024-06-28 at 1 09 11 PM

rauhul:
@swift-ci  test

rauhul:
@swift-ci test

dempseyatgithub:
From the screen shots, the contrast seems very low for the default in both light and dark mode.

Possibly a higher contrast border on the selected item would make it more clear which item is selected.

kubamracek:
This PR looks like moving into the right direction to me.

kaishin:
@rauhul As @dempseyatgithub asked, can we increase the contrast between active and inactive tabs? If you're unable to do it let me know and I'll make the change before merging.

assets/stylesheets/elements/_tabs.scss
```diff
@@ -0,0 +1,67 @@
...
+    padding: 0;
+
+    // "a" used on install page, "button" used on packages package
+    a, button {
+      background-color: var(--color-fill-secondary);
+      border: 1px solid var(--color-fill-tertiary);
+      border-radius: var(--border-radius);
+      color: var(--color-text);
+
+      // FIXME: Sink prefers-constrast lower into the color stack so each element doesn't need to
```
kaishin, line 15:
```suggestion
      // FIXME: Sink prefers-contrast lower into the color stack so each element doesn't need to
```

kaishin (commented):
Besides the contrast comment above, let's get the other PR this depends on merged first so it's easier to review the relevant changes.

kaishin (approved):
Looking better!

@rauhul When we spoke last time I mentioned a [contrast standard](https://www.wcag.com/designers/1-4-3-color-contrast/), but that only pertains to contrast between foregound and background colors, which I misremembered being the issue here. As far as active and inactive UI elements go, there is no standard per se, but a good practice is to ensure that the visual style of the selected element is different enough from the unselected ones. As with everything design, defining "enough" is not a science.

kaishin (commented):
@swift-ci test

kaishin:
@swift-ci test

PR #756, opened 2024-07-09
Clarify that package manifests must be located at the root of the repository in the Conceptual Overview section of the SwiftPM page

daveverwer:
### Motivation:

We recently had [an issue](https://github.com/SwiftPackageIndex/SwiftPackageIndex-Server/discussions/3197) raised with SPI where someone thought the rule that `Package.swift` manifests must be located at the root of the package repository was something that SPI was enforcing rather than a SwftPM requirement.

I thought (and subsequently double-checked) that it was a SwiftPM requirement for library packages, but my first step in those checks was to search the web and this page was the first official source that came up. It struck me that this would be a good place to make this requirement clear.

### Modifications:

Clarified the Creating a Library Package section of the SwiftPM page to make this clear.

### Result:

Screenshot 2024-07-15 at 13 09 04@2x

### DIFF
--- a/documentation/package-manager/_basic-usage.md
+++ b/documentation/package-manager/_basic-usage.md
@@ -58,7 +58,7 @@ example-package-playingcard
 └── Package.swift
 ~~~
 
-When creating a library package intended for use by other projects, the `Package.swift` manifest must reside at the top level/root of the package directory structure.
+When creating a library package intended for use as a dependency in other projects, the `Package.swift` manifest must reside at the top level/root of the package directory structure.
 
 Because the `PlayingCard` target does not produce an executable,
 it can be described as a _library_.
--- a/documentation/package-manager/_basic-usage.md
+++ b/documentation/package-manager/_basic-usage.md
@@ -58,6 +58,8 @@ example-package-playingcard
 └── Package.swift
 ~~~
 
+When creating a library package intended for use by other projects, the `Package.swift` manifest must reside at the top level/root of the package directory structure.
+
 Because the `PlayingCard` target does not produce an executable,
 it can be described as a _library_.
 A library is a target that builds a module which can be imported by other packages.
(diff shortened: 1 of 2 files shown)


0xTim (approved):

tayloraswift:
i don’t think this is universally correct, `Package.swift` can and often *does* occur in places that are not the root of a git repository. for example, benchmarks and examples often live in nested directories. it’s only the case that SwiftPM can use a package *as a dependency* if that package’s manifest lives at the root of the repository.

perhaps clarify that this only applies to packages that are intended to be used as version-controlled dependencies of other packages?

daveverwer:
> perhaps clarify that this only applies to packages that are intended to be used as version-controlled dependencies of other packages?

Thanks @tayloraswift, and of course you're correct that this is only for dependencies. I think given that , a better place for this might be in the "Creating a Library Package" section, around the file tree diagram. I just pushed an update to move it to there.

shahmishal:
cc: @bnbarham

documentation/package-manager/_basic-usage.md
```diff
@@ -58,6 +58,8 @@ example-package-playingcard
 └── Package.swift
 ~~~
 
+When creating a library package intended for use as a dependency in other projects, the `Package.swift` manifest must reside at the top level/root of the package directory structure.
+
```
bnbarham, line 62: IMO this would make slightly more sense to put after the next paragraph.

bnbarham (commented):
Thanks @daveverwer!

documentation/package-manager/_basic-usage.md
daveverwer: Thank you for the review @bnbarham! I moved it down by one paragraph. 👍

daveverwer:
@swift-ci please test

bnbarham (approved):

PR #791, opened 2024-08-20
Add base64 and cvs gems

federicobucchi:
### Motivation:

`csv` and `base64` are loaded from the standard library, but will no longer be part of the default gems in the future versions of Ruby.

### Modifications:

Add `csv` and `base64` gems

### Result:

Warnings disappear


federicobucchi:
@swift-ci please test

0xTim (approved):

PR #796, opened 2024-09-03
Fix SASS Deprecations

federicobucchi:
### Motivation:

Sass's behavior for declarations that appear after nested rules will be changing to match the behavior specified by CSS in an upcoming version. To keep the existing behavior, move the declaration above the nested rule. To opt into the new behavior, wrap the declaration in `& {}`.

### Modifications:

Fixed the deprecations above

### Result:

Deprecations are not there anymore


daveverwer (approved):

federicobucchi:
@swift-ci please test

kaishin:
Thanks for taking care of this!

PR #803, opened 2024-09-12
first step updating main contributing guide

parispittman:
https://github.com/swiftlang/swift/blob/main/CONTRIBUTING.md was recently created to make the majority of the contributing code content that speaks to majority /swift contributors out of the main guide 

follow up PRs will update the _contributing-code.md file removing dup text and updating with the /swift contributing link


ahoppen (approved):

dempseyatgithub:
With this change, the files `contributing/_llvm.md` and `contributing/_adding-library-dependencies.md` are not used anywhere else on the site.

I think the files should be removed as part of this PR instead of leaving them abandoned unless there is a plan to reuse them in the near future. (I think the content of these files is better suited for swiftlang/swift on GitHub than on Swift.org)

FYI, I have tagged the current commit as `info-architecture-baseline` so that there is a baseline to work from and we can make sure all existing content is accounted for, even with changes like this.

dempseyatgithub (changes requested):
With this change, the files contributing/_llvm.md and contributing/_adding-library-dependencies.md are not used anywhere else on the site.

I think the files should be removed as part of this PR instead of leaving them abandoned unless there is a plan to reuse them in the near future. (I think the content of these files is better suited for swiftlang/swift on GitHub than on Swift.org)

FYI, I have tagged the current commit as info-architecture-baseline so that there is a baseline to work from and we can make sure all existing content is accounted for, even with changes like this.

parispittman:
agreed @dempseyatgithub - good call out. I'll add those to this in a bit, ask for your review again, and then we'll merge 🥳

AnthonyLatsis (approved):

dempseyatgithub:
@swift-ci test

dempseyatgithub (approved):

PR #810, opened 2024-09-17
Improve multi-arch links detection by a generic solution and add Windows (arm64) to 6.0 release

stevapple:
This PR implements a generic way of detecting multiple archs for the latest release, and add Swift on Windows (arm64) to `swift_releases.yml`.

### Motivation:

`_includes/install/_build_release.md` was relying on a dedicated `include.aarch64` argument for generating an additional link. This is error-prone (eg. we may forget to specify the argument even if an aarch64 toolchain exists), not extensible (eg. if we want to add `riscv64` support we'll then need a new dedicated link) and not future proof (eg. if a new platform doesn't support `x86_64` we'll need a bunch of extra logic).

### Modifications:

`_includes/install/_build_release.md` was partially refactored that:
- drops the `aarch64` input argument;
- detects available archs directly from `swift_releases.yml`;
- removes all special treatment for x86_64 and use unified code for generating all download links.

`install/windows/index.md` was slightly updated to add similar handling for multi-arch.

`_data/builds/swift_releases.yml` was updated to add `arm64` to Swift 6.0 on Windows.

### Result:

Most changes happened internally. The only visible change is that Swift 6.0 toolchain will be available for Windows (arm64).

### DIFF
--- a/_data/builds/swift_releases.yml
+++ b/_data/builds/swift_releases.yml
@@ -1807,6 +1807,7 @@
       docker: Coming Soon #6.0-windowsservercore-ltsc2022
       archs:
         - x86_64
+        - arm64
     - name: Static SDK
       platform: static-sdk
       checksum: 7984c2cf175bde52ba6ea1fcbe27fc4a148a6237c41c719209c9288ed3ceb652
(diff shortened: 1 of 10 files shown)


shahmishal (approved):

shahmishal:
@swift-ci test

PR #815, opened 2024-09-18
Fix two H1 tags on language steering group page

alexandersandberg:
### Motivation:

There should only be a single H1 per page. It also applies the wrong hierarchical styling.

### Modifications:

- Changed the Charter H1 to an H2

### Result:

|Before|After
|---|---|
|CleanShot 2024-09-18 at 09 57 05|CleanShot 2024-09-18 at 09 57 25|


alexandersandberg:
@swift-ci please test

kaishin (approved):

PR #816, opened 2024-09-19
[cxx-interop] Document guarantees and assumptions for non-const C++ s…

fahadnayyar:
These assumptions and guarantees are important to be documented to make sure that when C++ Shared Reference Types are passed as a non-const parameter to a C++ function or method from swift, then user's don't make wrong assumptions about the default behaviour supported in the Swift compiler.

### DIFF
--- a/documentation/cxx-interop/index.md
+++ b/documentation/cxx-interop/index.md
@@ -1271,6 +1271,20 @@ owned/guaranteed calling conventions. The C++ callers must guarantee that `x` is
 Note that functions returning a shared reference type such as `returnSharedObject` transfer the ownership to the caller.
 The C++ caller of this function is responsible for releasing the object.
 
+If a C++ Shared Reference Type is passed as an argument to a C++ API from Swift, the Swift compiler *guarantees* that the passed value would be alive. Also Swift *assumes* that the C++ API is not consuming i.e., it returns with a valid object in the passed reference at the end of the C++ API call.
+
+```swift
+var obj = SharedObject.create()
+receiveSharedObject(obj) // Swift gaurantees that obj is alive
+```
+
+```c++
+void receiveSharedObject(SharedObject *sobj) {
+  ...
+  // Swift assumes that sobj is valid, non-null object at the end of this function
+}
+```
+
 ### Unsafe Reference Types
 
 The `SWIFT_UNSAFE_REFERENCE` annotation macro has the same effect as `SWIFT_IMMORTAL_REFERENCE`


fahadnayyar:
@swift-ci test

documentation/cxx-interop/index.md
```diff
@@ -1271,6 +1271,20 @@ owned/guaranteed calling conventions. The C++ callers must guarantee that `x` is
 Note that functions returning a shared reference type such as `returnSharedObject` transfer the ownership to the caller.
 The C++ caller of this function is responsible for releasing the object.
 
+If a C++ Shared Reference Type is passed as an argument to a C++ API from Swift, the Swift compiler *guarantees* that the passed value would be alive. Also Swift *assumes* that the C++ API is not consuming i.e., it returns with a valid object in the passed reference at the end of the C++ API call.
```
ravikandhadai, line 1274: @fahadnayyar  I would like to suggest these changes: 
>f a C++ Shared Reference Type is passed as an argument to a C++ API from Swift, the Swift compiler *guarantees* that the passed value would be alive.

Can we add the following after this sentence. 
... *guarantees* that the passed value would be alive, and retains ownership of the object. In other words, the argument is passed as +0 and there is no transfer of ownership. 

> Also Swift *assumes* that the C++ API is not consuming i.e., it returns with a valid object in the passed reference at the end of the C++ API call.

Can we rephrase this as follows:
The C++ function is responsible for ensuring that the object referred to by the parameter is alive during the call. The C++ function should not assume it has ownership of the object and should do necessary retain operations if it is needs to take ownership. 

If the argument is an *inout* (non-const reference), and if the C++ function overwrites the value with the new value, it is responsible for releasing the old object, and ensuring that the new object is properly retained so that Swift caller has ownership of the new object. Adhering to these rules are necessary to safely and correctly pass SWIFT_SHARED_REFERENCE between Swift and C++.  These rules are also general good practices and conventions to properly manage shared objects that use reference counting.

egorzhdan, line 1274: I have a few comments on the wording and formatting:

1. The rest of this document doesn't capitalize "C++ reference types", I think it's better to be consistent here.
2. I think it's better not to use bold highlighting for `*guarantees*`/`*assumes*`, the rest of the document doesn't use it, and I feel that it attracts too much attention.
3. Minor: the comma `,` usually goes before `i.e.`, could you please swap them?
4. I don't have a strong opinion, but to me "function call" sounds better than "API call", since to me API means a collection of functions, types, etc.

documentation/cxx-interop/index.md
```diff
@@ -1271,6 +1271,20 @@ owned/guaranteed calling conventions. The C++ callers must guarantee that `x` is
...
+
+```swift
+var obj = SharedObject.create()
+receiveSharedObject(obj) // Swift gaurantees that obj is alive
```
egorzhdan, line 1278: Typo: `guarantees`

documentation/cxx-interop/index.md
```diff
@@ -1271,6 +1271,20 @@ owned/guaranteed calling conventions. The C++ callers must guarantee that `x` is
...
+```
+
+```c++
```
egorzhdan, line 1281: Thanks for explicitly specifying the language for code blocks! 👍

shahmishal:
@egorzhdan @fahadnayyar Is this PR ready to be merged?

egorzhdan:
@shahmishal no, not yet.

hnrklssn:
Do we still plan to merge this?

documentation/cxx-interop/index.md
fahadnayyar: Thanks @egorzhdan . I've made these changes!

fahadnayyar: Thanks @ravikandhadai ! I've made these edits.

fahadnayyar:
@Xazax-hun @j-hui can you also please take a look at this patch?

documentation/cxx-interop/index.md
```diff
@@ -1280,6 +1280,32 @@ owned/guaranteed calling conventions. The C++ callers must guarantee that `x` is
 Note that functions returning a shared reference type such as `returnSharedObject` transfer the ownership to the caller.
 The C++ caller of this function is responsible for releasing the object.
 
+If a C++ shared reference type is passed as an non-const argument to a C++ API from Swift, the Swift compiler guarantees that the passed value would be alive, and retains the ownership of the value. In other words, the argument is passed as +0 and there is no transfer of ownership.
```
fahadnayyar: We should remove `non-const`.

documentation/cxx-interop/index.md
```diff
@@ -1255,6 +1255,34 @@ object.doSomething()
...
+
+```c++
+void receiveSharedObject(SharedObject *sobj) {
+  ...
+  // Swift assumes that sobj is valid, non-null object at the end of this function
+}
+```
+
+Note that if the argument is an inout (non-const reference) as shown below:
+```c++
```
j-hui, line 1281: There should be a blank line between the previous paragraph and the code block

```suggestion

```c++
```

documentation/cxx-interop/index.md
```diff
@@ -1255,6 +1255,34 @@ object.doSomething()
...
+
+```swift
+var obj = SharedObject.create()
+receiveSharedObject(obj) // Swift guarantees that obj is alive
+```
+
+```c++
+void receiveSharedObject(SharedObject *sobj) {
+  ...
+  // Swift assumes that sobj is valid, non-null object at the end of this function
```
j-hui, line 1270:
```suggestion
  sobj->doSomething() // sobj is valid until the end of receiveSharedObject()
```

documentation/cxx-interop/index.md
```diff
@@ -1255,6 +1255,34 @@ object.doSomething()
...
+}
+```
+
+Note that if the argument is an inout (non-const reference) as shown below:
+```c++
+void takeSharedObjectAsInout(SharedObject *& x) { ... }
+```
+
+which would be imported in Swift as
+```swift
```
j-hui, line 1287:
```suggestion

```swift
```

documentation/cxx-interop/index.md
```diff
@@ -1255,6 +1255,34 @@ object.doSomething()
...
+func takeSharedObjectAsInout(_ x: inout SharedObject) { ... }
+```
+
+ the C++ function can overwrite the value of the argument with the new value. However, the C++ function is responsible for releasing the old value, and ensuring that the new value is properly retained so that the Swift caller has ownership of the new value when the function returns. Adhering to these rules is necessary to safely and correctly pass around `SWIFT_SHARED_REFERENCE` between Swift and C++. These rules are also generally recommended conventions to manage shared objects that use reference counting.
```
j-hui:
```suggestion
The C++ function can overwrite the value of the argument with the new value. However, the C++ function is responsible for releasing the old value, and ensuring that the new value is properly retained so that the Swift caller has ownership of the new value when the function returns. Adhering to these rules is necessary to safely and correctly pass around `SWIFT_SHARED_REFERENCE` between Swift and C++. These rules are also generally recommended conventions to manage shared objects that use reference counting.
```

j-hui (commented):
A few small editorial nits, but haven't had time to closely the language yet

documentation/cxx-interop/index.md
```diff
@@ -1255,6 +1255,34 @@ object.doSomething()
 // `object` will be released here.
 ```
 
+### Calling conventions when passing shared reference types from Swift to C++
```
j-hui: I think this heading should be shorter:

```suggestion
### Passing Shared References from Swift to C++

```

Personally, I feel that the term "calling convention" is overloaded to encompass other aspects of ABI that are beyond the scope of this section (and should be mostly handled by the compiler/do not need to be user-facing for interop users). I think "ownership convention" is more semantically precise but I'm not sure if that is the right terminology to use in the context of Swift.

Also, headings should be followed by a blank line.

fahadnayyar: > Calling conventions when passing shared reference types from Swift to C++


@ravikandhadai suggested this heading, but I don't have strong feelings about it.

documentation/cxx-interop/index.md
```diff
@@ -1255,6 +1255,44 @@ object.doSomething()
...
+
+which would be imported in Swift as
+
+```swift
+func takeSharedObjectAsInout(_ x: inout SharedObject) { ... }
+```
+
+The C++ function can overwrite the value of the argument with the new value.
+However, the C++ function is responsible for releasing the old value, and ensuring that the new value is properly retained so that the Swift caller has ownership of the new value when the function returns.
+Adhering to these rules is necessary to safely and correctly pass around `SWIFT_SHARED_REFERENCE` between Swift and C++.
```
fahadnayyar, line 1293: @j-hui these 2 last lines are good, but I am not sure about their placement. Any suggestions?

Since we are editing this document with multiple calling conventions and rules about safely using SWIFT_SHARED_REFERENCE types and passing across the boundaries, I am not sure where to place these 2 lines.

Or maybe we can do such minor palcement edits later in a follow up patch?

documentation/cxx-interop/index.md
```diff
@@ -1255,6 +1255,44 @@ object.doSomething()
 // `object` will be released here.
 ```
 
+### Calling conventions when passing Shared Reference Types from Swift to C++
```
fahadnayyar, line 1258: TODO: ### -> ####

PR #828, opened 2024-10-11
Add placeholder API for swiftly releases

cmcgee1024:
As pitched in the forum this is a preliminary API providing metadata about the currently available swiftly release. https://forums.swift.org/t/new-swift-org-api-endpoint-for-the-current-swiftly-release/74702

### Motivation:

Swiftly itself has a self-update mechanism to help provide an easy path to keep in sync with the capability of the swift.org website, and to roll out new features and bug fixes. Since swift.org is the place where the official swiftly releases will be published then this API is updated whenever there is a new release. Since there has not yet been a swiftly release on swift.org and existing swiftly installations do not yet use this new API mechanism, there are no expected clients of this API yet. This is a preliminary step to prepare for a release of swiftly when that does happen.

### Modifications:

Added a new /api/v1/swiftly.json endpoint with metadata about the current swiftly release, for now, until the next swiftly is released.

### Result:

As a result, this endpoint will be accessible to swiftly when it is released.

### DIFF
--- a/api/v1/swiftly.json
+++ b/api/v1/swiftly.json
@@ -1,15 +1,4 @@
-{
-    "version": "0.3.0",
-    "platforms": [
-        {
-            "platform": "Darwin",
-            "x86_64": "https://download.swift.org/swiftly-0.3.0-release/darwin/swiftly-0.3.0-RELEASE/swiftly-0.3.0-RELEASE-darwin.pkg",
-            "arm64": "https://download.swift.org/swiftly-0.3.0-release/darwin/swiftly-0.3.0-RELEASE/swiftly-0.3.0-RELEASE-darwin.pkg"
-        },
-        {
-            "platform": "Linux",
-            "x86_64": "https://download.swift.org/swiftly-0.3.0-release/linux/swiftly-0.3.0-RELEASE/swiftly-0.3.0-RELEASE-linux.tar.gz",
-            "arm64": "https://download.swift.org/swiftly-0.3.0-release/linux-aarch64/swiftly-0.3.0-RELEASE/swiftly-0.3.0-RELEASE-linux-aarch64.tar.gz"
-        }
-    ]
-}
+---
+layout: none
+---
+{{ site.data.builds.swiftly_release | jsonify }}
--- a/api/v1/swiftly.json
+++ b/api/v1/swiftly.json
@@ -1,0 +1,15 @@
+{
+    "version": "0.3.0",
+    "platforms": [
+        {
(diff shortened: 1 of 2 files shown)


api/v1/swiftly.json
```diff
@@ -0,0 +1,15 @@
+{
+    "version": "0.3.0",
```
0xTim, line 2: Would we want to host more versions on here or would this always point to the latest?

0xTim:
Have we checked this get copied over automatically or do they need page metadata adding so Jekyll knows what to do with them?

api/v1/swiftly.json
cmcgee1024: In my current thinking, this would only ever have a single version of swiftly. I can't think of good reasons why someone should want to cherry-pick a particular version to install.

cmcgee1024:
> Have we checked this get copied over automatically or do they need page metadata adding so Jekyll knows what to do with them?

@0xTim the Jekyll mechanism is entirely opaque to me. I'm assuming a plain json without any template information will be taken literally by the template engine, but I'm not sure about that. To use the template engine for swiftly, I'd need to understand where in that data model the swiftly metadata will be kept.

Suggestions are certainly welcome from anyone who understands the system.

shahmishal:
We can do the same as `swift_releases.yml`, so we will need to create a new yaml file (swiftly_releases.yml) under `_data/build`.
for example: `swiftly_releases.yml`
```
- name: "0.3.0"
  tag: 0.3.0
  platforms:
    - name: Linux
      platform: Linux
      x86_64: "https://download.swift.org/swiftly-0.3.0-release/linux/swiftly-0.3.0-RELEASE/swiftly-0.3.0-RELEASE-linux.tar.gz"
```

After you will want to create a dir  `api/v1/swiftly`  and add a file `releases.json`
```
---
layout: none
---
{{ site.data.builds.swiftly_releases | jsonify }}

```

cmcgee1024:
@0xTim @shahmishal thank you, I've refactored the changes as you've suggested.

0xTim:
@swift-ci test

0xTim:
@swift-ci please test

shahmishal:
Moving this to draft until swiftly tar.gz / pkg are available on download.swift.org

czechboy0:
Related to https://github.com/swiftlang/swift-org-website/pull/841, depending on which one lands first, we'll want to update the other one.

federicobucchi (approved):

federicobucchi:
@swift-ci please test

PR #837, opened 2024-10-26
Story book improvements

kaishin:
- Introduce a Jekyll plugin for sample code.
- Remove list of components at the top to avoid manual update requirement.
- Simplify the samples.
- Use rem for measurements.
- Move around styles to better match their content.
- Other minor changes.


federicobucchi (approved):
LGTM

PR #846, opened 2024-11-07
Streamline Windows installer page

### LINKED ISSUE #845: Adjust "Traditional" Install Instructions for Swift on Windows
I wanted to start a quick conversation about what we should do with the "traditional" installation instructions for Windows. 
https://www.swift.org/install/windows/traditional/

The instructions are designed around installing Swift 5.4.2, with unnecessary or obsolete steps for modern Swift installations. I don't think we want to encourage people to install such an old version. At the same time, I think we want to have a basic README associated with a manual installation: in particular its dependencies.

We could optionally keep an archived page for Swift 5.4 and older, but my sense is that this is unnecessary. It's an anti-pattern to install those versions, and I don't even believe we publish these downloads anywhere (if we do, I can't find them!)

My proposal is that we clean this page up and focus it on modern (Swift 6+) installs, removing unnecessary steps like GPG key validation  and targeting a more advanced user (i.e. note the dependencies, don't necessarily provide handholding of how to install them). Curious on other thoughts from @compnerd and others.

timsneath:
Simplify manual installation instructions for Windows by:

- Creating a streamlined instructions page for installing Swift 6 on Windows without a package manager
- Relocating the laborious instructions for older (5.9 and earlier) Swift releases to a separate page
- Updating the top-level Windows installer page to include links to the new pages

I've also reviewed the instructions for modern Swift builds and:
- Harmonized them with the instructions for package managers (e.g. don't require CMake Project Tools)
- Changed the dependencies for Python and Git to point to the source locations, rather than the vended versions in Visual Studio (since Python in particular is no longer vended)

Fixes #845


timsneath:
@swift-ci please test

install/windows/_manual.md
```diff
@@ -0,0 +1,34 @@
...
+| MSVC v143 - VS 2022 C++ ARM64/ARM64EC build tools (Latest)[^1] | Microsoft.VisualStudio.Component.VC.Tools.ARM64 |
+| Windows 11 SDK (10.0.22000.0)[^2] | Microsoft.VisualStudio.Component.Windows11SDK.22000 |
+
+[^1]: At minimum, Swift requires the build tools that match your machine architecture. Installing other architectures is recommended in order to cross-compile Swift binaries to run on different machine architectures.
+
+[^2]: You may install a newer SDK instead.
+
+You should also install the following dependencies:
+
+- [Python 3 (latest stable release)](https://www.python.org/downloads/windows/)
```
compnerd, line 19: The latest version is not recommended. You should be using 3.10.x as that is what we build against and because of the way that LLDB uses Python, we need the version match.

install/windows/_manual.md
```diff
@@ -0,0 +1,34 @@
...
+
+Developer Mode enables debugging and other settings that are necessary for Swift development. Please see Microsoft’s [documentation](https://docs.microsoft.com/windows/apps/get-started/enable-your-device-for-development) for instructions about how to enable developer mode.
+
+## Install Swift
+
+After the above dependencies have been installed, [download and run the installer for the latest Swift stable release ({{ site.data.builds.swift_releases.last.name }}](/install/windows)). 
+
+Alternatively, you may prefer to install a [development snapshot](/install/windows/#development-snapshots) for access to features that are actively under development. 
+
+The toolchain on Windows is typically installed to `%SystemDrive%\Library\Developer\Toolchains`.
```
compnerd, line 32: This path has now changed with 5.9 I think? We install to a per-user install location (`%LocalAppData%\Programs\Swift\Toolchains`).

install/windows/manual/index.md
```diff
@@ -0,0 +1,39 @@
...
+| MSVC v143 - VS 2022 C++ ARM64/ARM64EC build tools (Latest)[^1] | Microsoft.VisualStudio.Component.VC.Tools.ARM64 |
+| Windows 11 SDK (10.0.22000.0)[^2] | Microsoft.VisualStudio.Component.Windows11SDK.22000 |
+
+[^1]: At minimum, Swift requires the build tools that match your machine architecture. Installing other architectures is recommended in order to cross-compile Swift binaries to run on different machine architectures.
+
+[^2]: You may install a newer SDK instead.
+
+You should also install the following dependencies:
+
+- [Python 3 (latest stable release)](https://www.python.org/downloads/windows/)
```
compnerd, line 24: Same as above

install/windows/manual/index.md
```diff
@@ -0,0 +1,39 @@
...
+
+Developer Mode enables debugging and other settings that are necessary for Swift development. Please see Microsoft’s [documentation](https://docs.microsoft.com/windows/apps/get-started/enable-your-device-for-development) for instructions about how to enable developer mode.
+
+## Install Swift
+
+After the above dependencies have been installed, [download and run the installer for the latest Swift stable release ({{ site.data.builds.swift_releases.last.name }}](/install/windows)). 
+
+Alternatively, you may prefer to install a [development snapshot](/install/windows/#development-snapshots) for access to features that are actively under development. 
+
+The toolchain on Windows is typically installed to `%SystemDrive%\Library\Developer\Toolchains`.
```
compnerd, line 37: And here too

install/windows/_manual.md
timsneath: Good catch -- thanks!

timsneath: Good catch -- thanks! When does this change, though? If I put a static version in, won't it become quickly stale (the original

compnerd: Yes, it does run the risk of becoming stale. Unfortunately, I don't think that there is a good solution here, unless you want to try to parse [build.ps1](https://github.com/swiftlang/swift/blob/main/utils/build.ps1).

This changes somewhat infrequently - I am trying to match the major/minor version that Apple has on their OS (i.e. 3.10). You can use the latest patch release for that version and it should be fine.

timsneath:
@compnerd Thanks for your feedback; please take another look and see if this addresses your comments. In particular, I've  added a footnote to include your explanation on which Python version should be selected. 

@swift-ci please test

compnerd (approved):
Perhaps use `archived` instead of `old`? But if that doesn't show up in the URL, I don't think that it really matters at all.

timsneath:
@swift-ci please test again. thanks for all you do behind the scenes, my little unsung hero 🦸

install/windows/_manual.md
timsneath: @compnerd Actually I have more questions here, if you'll indulge me. You said that build.ps1 shows the version that's used to build Swift -- but the [current version there](https://github.com/swiftlang/swift/blob/fb0b3fa90c72ae9de6a17d97097e706298ff2d02/utils/build.ps1#L130) is 3.9.10, not 3.10.x. Did I miss something?

And regardless of whether it's 3.9 or 3.10, what do we expect an ARM64 user to do? As far as I can tell, the first ARM64 version supported is [3.11.0](https://www.python.org/downloads/release/python-3110/), and I wasn't able to get AMD64 builds to successfully install in emulation mode.

compnerd: Ah, sorry, I must've misremembered the version. For ARM64, we currently pull a version from nuget. This is the only way that we have to avoid a full build of python as a dependency.

As to the criticality of it - without the specific version being installed and on the path, lldb will not function. This extends to the REPL, though that is currently known to have a regression with 6.x. The missing functionality of lldb will extend to any IDE (e.g. VSCode).

Given that LLDB is part of the expected experience for Swift, I would consider it pretty critical.

I am reluctant to move the dependency independently of the other platforms (i.e. Apple). At the same time, even if we were to move the dependency independently, we wouldn't see that in practice until swift next. For the time being it would appear that we are stuck with this version.

shahmishal:
@swift-ci please test

timsneath:
@swift-ci please test

shahmishal (approved):

PR #850, opened 2024-11-18
Add SSWG package collections

finestructure:
### Motivation:

This adds the data files to manage packages following the [SSWG incubation process](https://www.swift.org/sswg/incubation-process.html). They will also be the basis for custom package collections on the Swift Package Index.

### Modifications:

Adds three files with package URLs.

To do, as discussed with @0xTim :

- [ ] make sure the YML files are configured to be processed to JSON files at the appropriate URL, probably `https://github.com/swiftlang/swift-org-website/blob/main/api/v1/install/{graduated|incubating|sandbox}.json` - perhaps they should have an `sswg-prefix`?
- [ ] update the files to reflect the latest review

### DIFF
--- a/_data/server-workgroup/graduated.yml
+++ b/_data/server-workgroup/graduated.yml
@@ -1,0 +1,12 @@
+- https://github.com/apple/swift-crypto.git
+- https://github.com/apple/swift-log.git
+- https://github.com/apple/swift-metrics.git
+- https://github.com/apple/swift-nio.git
+- https://github.com/apple/swift-statsd-client.git
+- https://github.com/grpc/grpc-swift.git
+- https://github.com/swift-server-community/APNSwift.git
+- https://github.com/swift-server/async-http-client.git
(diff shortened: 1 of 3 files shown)


0xTim:
Do you think it's worth combining these into one file? Would be easier to spot duplicate and changes etc and I think the tools should be fine

finestructure:
Could do that but it would definitely have to _produce_ three separate files for the package collections - and that's probably not straightforward, is it?

daveverwer:
> Nope we should be able to read a subsection of the YAML easily. I'll have a play

One YAML -> multiple output files can be done with a Jekyll generator. Take a look at what I did in `_plugins/packages.rb` for how to do it.

0xTim:
@finestructure https://github.com/swiftlang/swift-org-website/blob/main/_data/server-workgroup/projects.yml 😆

0xTim:
Closing in favour of #853

PR #862, opened 2024-12-16
Rework homepage / getting started flow

timsneath:
Bring getting started content front and center, streamlining the homepage and focusing on different Swift use cases.

### Motivation:

Today's home page focuses on just three scenarios: Apple platforms, CLI and server. A separate "Getting Started" page duplicates some of these and adds other choices (e.g. embedded). This is confusing. Additionally, the first steps for installation vary depending on the scenario (e.g. a CLI developer may want the static toolchain, an app developer wants Xcode, a cloud developer wants Docker, and an embedded developer needs cross-compilation instructions. 

The goal is to get any new Swift user to a great first hour experience focused on their needs. 

### Modifications:

The home page will contain a catalog of all scenarios that a developer can use Swift for. And for each of these, a separate page will provide some simple steps, a small sample of code, and a graphic of Swift running in that scenario, along with a list of packages that will be helpful, demonstrating the breadth of ecosystem for each. We'll do that by:

- Revising the "Use Cases" section into a full list of all the wonderful things you can do with Swift
- Removing the separate "Getting Started" page from the navbar, streamlining to one "install" call-to-action
- Adding new getting started pages that are focused on the first hour experience for each scenario

### Result:

Desired result is more developers clicking on one of the scenarios, and getting to a delightful first hour with Swift that is tailored around their needs.


getting-started/cli-swiftpm/index.md
```diff
@@ -3,19 +3,16 @@ layout: page
 title: Build a Command-line Tool
 ---
 
-> The source code for this guide can be found [on GitHub](https://github.com/apple/swift-getting-started-cli)
```
0xTim, line 6: Why have we removed this link to the repo?

getting-started/cli-swiftpm/index.md
```diff
@@ -141,32 +138,34 @@ let package = Package(
...
+        Figlet.say(input)
+    }
 }
 ~~~
 
 For more information about how [swift-argument-parser](https://github.com/apple/swift-argument-parser) parses command line options, see [swift-argument-parser documentation](https://github.com/apple/swift-argument-parser) documentation.
 
-Once we save that, we can run our application with `swift run MyCLI --input 'Hello, world!'`
+Once we save that, we can run our application with `swift run MyCLI --input
+'Hello, world!'` (macOS / Linux) or `swift run MyCLI --input "Hello, world!"`
```
0xTim, line 167: This line break will break the monospaced formatting

getting-started/cli-swiftpm/index.md
```diff
@@ -31,53 +28,53 @@ This will generate a new directory called MyCLI with the following files:
...
+// swift-tools-version: 6.0
 // The swift-tools-version declares the minimum version of Swift required to build this package.
 
 import PackageDescription
 
 let package = Package(
     name: "MyCLI",
     dependencies: [
-      .package(url: "https://github.com/apple/example-package-figlet", branch: "main"),
+        .package(url: "https://github.com/apple/example-package-figlet", branch: "main")
```
0xTim, line 62:
```suggestion
        .package(url: "https://github.com/apple/example-package-figlet", branch: "main"),
```

getting-started/cli-swiftpm/index.md
```diff
@@ -31,53 +28,53 @@ This will generate a new directory called MyCLI with the following files:
 
 `Sources/main.swift` is the application entry point and where we’ll write our application code.
 
-In fact, SwiftPM generated a "Hello, world!" project for us!
+In fact, Swift generated a "Hello, world!" project for us!
```
0xTim, line 31: Are we happy with this? Because this is technically incorrect, it's SwiftPM that generates the project

getting-started/swiftui/index.md
```diff
@@ -1,413 +1,63 @@
...
-Go ahead and launch Xcode once it’s installed, then select Create a new Xcode Project. Select the iOS tab at the top, then select the App template and press Next.
+When building mobile apps with Swift, you'll want a full development
+environment, rather than just the Swift compiler. Xcode provides the latest
+version of Swift, along with tools for designing and previewing your user
+interface, simulators for testing your app, and tools for uploading it to the
+App Store.
 
-**Tip:** Although we’ll be targeting iOS 16, our code will also work great on macOS Ventura and beyond.
+Xcode is designed for running on Mac, and you can download it for free from [the
+Mac App Store](https://apps.apple.com/us/app/xcode/id497799835).
```
0xTim, line 22: I think we've discussed this in the past, but we should link to https://developer.apple.com/download/applications/ instead. Mac App Store links don't work everywhere and are confusing on mobile

getting-started/swiftui/tutorial.md
```diff
@@ -0,0 +1,409 @@
+---
+layout: page
+title: Build Mobile Apps with Swift
+---
+
+In this tutorial you’re going to use Swift and SwiftUI to build a small app to recommend fun new activities to users. Along the way you’ll meet several of the basic components of a SwiftUI app, including text, images, buttons, shapes, stacks, and program state.
+
+To get started, you’ll need to [download Xcode from the Mac App Store](https://apps.apple.com/app/xcode/id497799835?mt=12). It’s free, and comes with Swift and all the other tools you need to follow this tutorial.
```
0xTim, line 8: See above for comment about linking to the App Store

getting-started/cli-swiftpm/index.md
timsneath: Yeah, this should probably stay. It feels duplicative to have it at the top _and_ the bottom, though -- what do you think about removing the bottom reference? 

(The rationale was to reduce extraneous content from these guides to get to a short punchy intro to each area.)

getting-started/cli-swiftpm/index.md
timsneath: Looks OK to me? 
image

getting-started/cli-swiftpm/index.md
timsneath: My rationale is simply that SwiftPM is an integral part of Swift. That the `swift` command line utility calls SwiftPM to do the work of creating the template is an implementation detail during the first hour of experiencing Swift, in the same way that we don't need to say: "Running swift build will instruct _SwiftPM_ to download the new dependencies and then proceed to build the code" even though your argument applies there too.

Thoughts? We don't want to be incorrect, obviously :)

getting-started/swiftui/index.md
timsneath: Double-checking w/ the Xcode team.

timsneath: Feedback from Xcode: they strongly prefer to see customers directed to the App Store because of its streamlined, auto-updated installation. FWIW, I tried the link on an iPhone, and it loads the App Store successfully, presenting a reasonable screen with a warning that it can't be installed locally. 
98941B56-A663-4001-B5E2-720B8285E939_4_5005_c

0xTim: I'm not overly keen on it (auto updates are problematic and slow, the App Store still isn't on 16.2, the developer site links to both Mac App Store and direct download etc but it's not something I feel super strong about

getting-started/cli-swiftpm/index.md
0xTim: They were added at top and bottom to make it easy to find the code as that's what many people want to look at, either right at the start or when finishing the tutorial. I don't think it adds too much bloat having it in both

getting-started/cli-swiftpm/index.md
0xTim: Fine by me

getting-started/swiftui/index.md
dempseyatgithub: There also have been a number of time over the past years where after a new Xcode release, the direct download has been fine but there has been something wrong with the Mac App Store version.

I do use the Mac App Store version on my own machine but I always wait a week or so before updating because this has happened regularly enough to make me wary of updating immediately.

PR #873, opened 2025-01-18
Use swiftly_release data for the download page

cmcgee1024:
Change the download pages for Linux and macOS so that they use the swiftly_release information to get the latest swiftly version to set the URL's and instructions.

### Motivation:

We would like to use the API to automatically populate the swiftly information when a new release is made instead of having to update the Linux and macOS pages each time.

### Modifications:

Pulled in the latest commits from main with the swiftly release API.

### Result:

The change will automatically set the URLs and instructions relating to the swiftly version.


0xTim:
Is there a way to split out the Swiftly stuff from the updates from `main`?

0xTim:
@cmcgee1024 want to fix the merge conflicts? Then we can merge

PR #875, opened 2025-01-20
[Server Guides] Logging clarifications

czechboy0:
### Motivation:

More guidance was needed on whether libraries should mutate Logger's log level and log handler or not.

### Modifications:

Clarify that libraries should not do that, with more context. Also removed language using "higher" and "lower" for log levels, as it's ambiguous, and replaced with "more severe" and "less severe".

### Result:

Clearer documentation for Swift server library authors around logging.


FranzBusch (approved):

ktoso (approved):
Nice clarifications, thanks Honza. Lgtm !

0xTim (approved):

0xTim:
@swift-ci test

PR #879, opened 2025-01-24
Clarify that `.pkg` installer is required for cross-compilation

MaxDesiatov:
### Motivation:

When cross-compiling on macOS to non-Darwin platforms (Linux or Embedded Swift), `.pkg` installer is required, as it provides necessary tools and has more LLVM backends enabled.

### Modifications:

https://www.swift.org/install/macos/ page now clarifies which installation methods are suitable for each of the use cases.

### Result:

There's less confusion about versions of toolchains necessary for cross-compilation to non-Darwin platforms.

### DIFF
--- a/install/macos/index.md
+++ b/install/macos/index.md
@@ -11,14 +11,14 @@ title: Install Swift
   <li class="grid-level-1">
     <h3>Xcode</h3>
     <p class="description">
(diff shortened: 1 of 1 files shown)


0xTim:
@swift-ci please test

PR #880, opened 2025-01-24
Fix REPL and build instructions in `install/linux/tarball`

MaxDesiatov:
The instructions provided are incorrect: in recent Swift versions the `swift` command won't run the REPL and users need to execute `swift repl` explicitly. Additionally, on non-Darwin platforms users build Swift packages not "Swift projects", the `swift build` command should be utilized for that.


0xTim (approved):

0xTim:
@swift-ci please test

PR #886, opened 2025-02-02
Add GH Action for CI

0xTim:
First steps for adopting GH actions with an action to run on PRs to make sure the site works. This runs:

* the soundness script (which has already picked up an error)
* `jekyll build`
* The OpenAPI test client to ensure the OpenAPI spec works (cc @czechboy0)

@shahmishal can we can actions enabled? I don't know what the original CI was doing, I'm assuming just a Jekyll build

### DIFF
--- a/.github/workflows/ci.yaml
+++ b/.github/workflows/ci.yaml
@@ -37,6 +37,6 @@ jobs:
           DOWNLOADSWIFTORG_SERVER_NAME: local
         run: |
             docker compose -f ../../docker-compose.yaml run build 
-            docker compose -f ../../docker-compose.yaml start website
+            docker compose -f ../../docker-compose.yaml run -d -p 4000:4000 website
             swift run swiftorgClient
             swift run downloadswiftorgClient
--- a/.github/workflows/ci.yaml
+++ b/.github/workflows/ci.yaml
@@ -33,7 +33,10 @@ jobs:
         uses: actions/checkout@v4
       - name: Run OpenAPI Tester
         working-directory: openapi/TestSwiftOrgClient
+        env:
+          DOWNLOADSWIFTORG_SERVER_NAME: local
         run: |
             docker build -t swift-org-site ../../. 
             docker run -p 4000:4000 -d --name website swift-org-site
-            swift run
+            swift run swiftorgClient
+            swift run downloadswiftorgClient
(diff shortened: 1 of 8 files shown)


.github/workflows/ci.yaml
```diff
@@ -0,0 +1,37 @@
...
+      - name: Build site
+        run: bundle install && bundle exec jekyll build
+
+  test-openapi:
+    runs-on: ubuntu-latest
+    container: swift:6.0
+    steps:
+      - name: Check out
+        uses: actions/checkout@v4
+      - name: Run OpenAPI Tester
```
czechboy0, line 35: If we want to test that the JSON files backing the API actually still work, we'll need to start the jekyll server locally, and then tell the test CLI to fetch from that.

This might require installing Swift in the container based on the jekyll image, here in the pipeline. Or install jekyll into the Swift one.

czechboy0 (commented):
Thanks @0xTim! 🙏

.github/workflows/ci.yaml
0xTim: Ahh ok. I've updated it so that we build the container and run that - also helps us test the container image is valid.

Since the default runner has Swift installed and Docker we shouldn't need to run it in a container

shahmishal (approved):
Thanks

.github/workflows/ci.yaml
```diff
@@ -0,0 +1,43 @@
...
+
+  test-openapi:
+    runs-on: ubuntu-latest
+    steps:
+      - name: Check out
+        uses: actions/checkout@v4
+      - name: Run OpenAPI Tester
+        working-directory: openapi/TestSwiftOrgClient
+        env:
+          DOWNLOADSWIFTORG_SERVER_NAME: local
```
czechboy0, line 37: And SWIFTORG_SERVER_NAME=local is the other one

0xTim: Just beat you 😉

czechboy0: Nvm you were faster and already did it 🫡

czechboy0 (approved):

czechboy0:
Not sure if the evolution API (downloads.swift.org) runs on the local jekyll server, actually.

0xTim:
@shahmishal can you remove the required mark for the old website build and switch it to the new one then merge?

PR #913, opened 2025-02-21
Update the footer and the home page to include Bluesky and Mastodon a…

shahmishal:
…ccount links

https://github.com/swiftlang/swift-org-website/issues/912

Screenshot 2025-02-21 at 11 54 11 AM


_includes/footer.html
```diff
@@ -31,7 +31,9 @@
       </label>
     </form>
     <aside>
-      <a href="https://twitter.com/swiftlang" rel="nofollow" title="Follow @SwiftLang on Twitter"><i class="twitter"></i></a>
+      <a href="https://x.com/swiftlang" rel="nofollow" title="Follow @SwiftLang on X"><i class="x"></i></a>
+      <a href="https://bsky.app/profile/swift.org" rel="nofollow" title="Follow @swift.org on Bluesky"><i class="bluesky"></i></a>
+      <a href="https://mastodon.social/@swiftlang" rel="nofollow" title="Follow @swiftLang on Mastodon"><i class="mastodon"></i></a>
```
timsneath, line 36: Do we want `rel="me"` rather than `rel="nofollow"` for these links? [Reading around](https://developer.mozilla.org/en-US/docs/Web/HTML/Attributes/rel/me) it's a) required for Mastodon verification, and b) seems like the right semantics for our purpose.

shahmishal: Updated

PR #933, opened 2025-03-21
Permit freeform source branch and platform identifiers in OpenAPI spec

cmcgee1024:
When using the current OpenAPI specification for discovering snapshot toolchains
in an OpenAPI generator, there is a code-level limitation on the known branches
(e.g. main or 6.0), and the known platform identifiers (e.g. macOS, amazonlinux2).
These values will change over time and new snapshot branches come (e.g. 6.1, or 7.0)
and platforms are added (e.g. "ubuntu2404"). Clients that are based on earlier versions
of the schema should remain compatible, and permit user input of new values that they
know to exist. That's not possible currently.

This change will adjust the schema to take advantage of the schema types that allow a
choice of a known value, and ones that are not (yet) known to the schema.

### DIFF
--- a/openapi/swiftorg.yaml
+++ b/openapi/swiftorg.yaml
@@ -37,12 +37,12 @@ paths:
         in: path
         required: true
         schema:
-          $ref: '#/components/schemas/KnownSourceBranch'
+          $ref: '#/components/schemas/SourceBranch'
       - name: platform
         in: path
         required: true
         schema:
-          $ref: '#/components/schemas/KnownPlatformIdentifier'
+          $ref: '#/components/schemas/PlatformIdentifier'
     get:
       operationId: listDevToolchains
       summary: Fetch all development toolchains
@@ -61,7 +61,7 @@ paths:
         in: path
         required: true
         schema:
-          $ref: '#/components/schemas/KnownSourceBranch'
+          $ref: '#/components/schemas/SourceBranch'
     get:
       operationId: listStaticSDKDevToolchains
       summary: Fetch all static SDK development toolchains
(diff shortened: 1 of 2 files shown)


0xTim:
@czechboy0

openapi/swiftorg.yaml
```diff
@@ -37,12 +37,12 @@ paths:
         in: path
         required: true
         schema:
-          $ref: '#/components/schemas/KnownSourceBranch'
+          $ref: '#/components/schemas/SourceBranch'
```
czechboy0, line 40: It was a deliberate decision to make the parameters have to be one of the known values, and in responses to return a flexible extensible value which might or might not be known.

Is this because swiftly might get _ahead_ of the OpenAPI doc, and want to provide a case that's not known to swift.org?

cmcgee1024: it's actually more of a case that swiftly could lag behind in its understanding of the spec. A user might still want to install a nightly toolchain from what is the current branch.

There is also a desire to make swiftly able to automatically handle new Linux platforms, since the binary is compiled with the static Linux SDK, so from an executable perspective it can run on virtually any Linux distribution. As long as the swift.org API supports it then swiftly can provide support. We're not there today, but that's something that's been discussed as an enhancement to the API.

czechboy0: > able to automatically handle new Linux platforms

That's totally fair, and a requirement I didn't know about. Makes total sense, let's update the parameters to be open enums rather than closed enums then.

czechboy0 (approved):
Thanks, @cmcgee1024

0xTim (approved):

PR #944, opened 2025-03-29
[Swiftly] Download the swiftly package in /tmp/

shahmishal:
Screenshot 2025-03-28 at 10 03 01 PM
Screenshot 2025-03-28 at 10 03 13 PM


0xTim:
This can cause issues on some Linux deployments where `/tmp` is just mounted as local memory (e.g. RAM) and there's not enough space to download and unzip it. The current behaviour also matches tools such as rustup

PR #950, opened 2025-03-29
Replace deprecated `docker-compose` with `docker compose` in the instructions

kkebo:
Since Compose V2, `docker-compose` has been deprecated, and instead, `docker compose` has been recommended.

https://docs.docker.com/compose/releases/migrate/#docker-compose-vs-docker-compose


0xTim (approved):
Thanks!

PR #958, opened 2025-04-04
Document features around safe interoperability

Xazax-hun:
This PR adds a new top-level page describing how to interact with C++ from strict memory safe Swift.

### DIFF
--- a/documentation/cxx-interop/safe-interop/index.md
+++ b/documentation/cxx-interop/safe-interop/index.md
@@ -1,0 +1,297 @@
+---
+layout: page
+title: Mixing Strict Memory Safe Swift and C++
+official_url: https://swift.org/documentation/cxx-interop/safe-interop/
+redirect_from:
+- /documentation/cxx-interop/safe-interop.html
+---
+
+## Table of Contents
+{:.no_toc}
+
+* TOC
+{:toc}
+
+## Introduction
+
+[Strict memory safe Swift](https://github.com/swiftlang/swift-evolution/blob/main/proposals/0458-strict-memory-safety.md)
+is a new feature in Swift 6.2 to make code easier to audit for memory safety.
+This document describes how to ergonomically interact with Swift by importing
+C++ construct safely. All of the features in this document work in regular Swift
+but they provide more value in strict memory safe mode.
+
+* * *
+
+<div class="info" markdown="1">
+C++ interoperability is an actively evolving feature of Swift.
+Future releases of Swift might change how Swift and C++
+interoperate,
+as the Swift community gathers feedback from real world adoption of C++
+interoperability in mixed Swift and C++ codebases.
+Please provide the feedback that you have on the
+[Swift forums](https://forums.swift.org/c/development/c-interoperability/), or
+by filing an [issue on GitHub](https://github.com/swiftlang/swift/issues/new/choose).
+Future changes to the design or functionality of C++ interoperability will not
+break code in existing codebases [by default](#source-stability-guarantees-for-mixed-language-codebases).
+</div>
+
+## Overview
+
+Swift provides memory safety with a combination of language affordances and runtime checking.
+However, Swift also deliberately includes some unsafe constructs, such as the Unsafe pointer types in the standard library.
+Swift occasionally needs additional information that is not present in the C++ type and API declarations
+to safely interface with them. This document describes how such code need to be annotated.
+
+### Annotating foreign types
+
+Types imported from C++ are considered foreign to Swift. Many of these types are considered safe,
+including the built-in integral types like `int`, some standard library types like `std::string`,
+and aggregate types that are using safe types are building blocks. 
+
+On the other hand, some C++ types are imported as unsafe by default. Consider the following C++ type
+and APIs:
+
+```c++
+class StringRef {
+public:
+    StringRef() : ptr(nullptr), len(0) {}
+    
+    std::string toString() const;
+private:
+    const char* ptr;
+    size_t len;
+};
+
+std::string normalize(const std::string& path);
+
+// The path needs to be normalized.
+StringRef fileName(const std::string& path);
+```
+
+Let's try to use them from Swift with strict memory safety enabled:
+
+```swift
+func getFileName(_ path: borrowing std.string) -> StringRef {
+    let normalizedPath = normalize(path)
+    return fileName(normalizedPath)
+}
+```
+
+Building this code will emit a warning that the `fileName` call is unsafe because
+it references the unsafe type `StringRef`. Swift considers `StringRef` unsafe because
+it has a pointer member. Types like `StringRef` can dangle, we need to take extra
+care using them, making sure the referenced buffer outlives the `StringRef` object.
+
+Swift's [non-escapable types](https://github.com/swiftlang/swift-evolution/blob/main/proposals/0446-non-escapable.md)
+can also have lifetime dependencies, just like `StringRef`. However, the Swift compiler
+can track these dependencies and enforce safety at compile time. To import `StringRef`
+as a safe type we need to mark it as not escapable.
+
+```c++
+class SWIFT_NONESCAPABLE StringRef {
+```
+
+### Annotating APIs
+
+Building the code again will emit a new diagnostic for the `fileName` function about
+missing lifetime annotations. Functions returning non-escapable types need annotations
+to describe their lifetime contracts via [lifetimebound](https://clang.llvm.org/docs/AttributeReference.html#id11)
+and [lifetime_capture_by](https://clang.llvm.org/docs/AttributeReference.html#lifetime-capture-by) annotations.
+
+```c++
+StringRef fileName(const std::string& path [[clang::lifetimebound]]);
+```
+
+This annotation describes that the returned `StringRef` value has the same lifetime as the argument
+of the `fileName` function.
+
+Building the project again reveals a lifetime error in the Swift function:
+
+```swift
+func getFileName(_ path: borrowing std.string) -> StringRef {
+    let normalizedPath = normalize(path)
+    // error: lifetime-dependent value escapes local scope
+    // note: depends on `normalizedPath`
+    return fileName(normalizedPath)
+}
+```
+
+The value returned by `fileName` will dangle after the lifetime of `normalizedPath` ends.
+We can fix this error by pushing the task of normalizing a path to the callee:
+
+```swift
+// Path needs to be normalized.
+func getFileName(_ path: borrowing std.string) -> StringRef {
+    return fileName(normalizedPath)
+}
+```
+
+Or we could return a self-contained value like `std.string` instead of a dangling `StringRef`:
+
+```swift
+func getFileName(_ path: borrowing std.string) -> std.string {
+    let normalizedPath = normalize(path)
+    let ref = fileName(normalizedPath)
+    return ref.toString()
+}
+```
+
+## Escapability annotations in detail
+
+All of the unannotated types are imported as `Escapable` to maintain backward
+compatibility. This might change in the future under a new interoperability version.
+We can import a type as `~Escapable` to Swift by adding the `SWIFT_NONESCAPABLE` 
+annotation.
+
+```c++
+struct SWIFT_NONESCAPABLE View {
+    View() : member(nullptr) {}
+    View(const int *p) : member(p) {}
+    View(const View&) = default;
+private:
+    const int *member;
+};
+```
+
+Moreover, we can explicitly mark types as `Escapable` using the `SWIFT_ESCAPABLE`
+annotation.
+
+```c++
+struct SWIFT_ESCAPABLE Owner {
+    ...
+}; 
+```
+
+The main reason for explicitly annotating a type as `SWIFT_ESCAPABLE` to make sure
+it is considered as a safe type when used from Swift. Functions returning escapable
+types do not need lifetime annotations.
+
+Escapability annotations can be attached to types via APINotes.
+
+```
+Tags:
+- Name: NonEscapableType
+  SwiftEscapable: false
+- Name: EscapableType
+  SwiftEscapable: true
+```
+
+In case of template instantiations the escapability of a type can depend on the
+template arguments:
+
+```c++
+MyList<View> f();
+MyList<Owner> g();
+```
+
+In this example, `MyList<View>` should be imported as `~Escapable` while `MyList<Owner>`
+should be imported as `Escapable`. This can be achieved via conditional escapability
+annotations.
+
+```
+template<typename T>
+struct SWIFT_ESCAPABLE_IF(T) MyList {
+    ...
+};
+```
+
+## Lifetime annotations in detail
+
+The `lifetimebound` annotation can be used to annotate methods, free functions, and constructors.
+It can annotate dependence on regular arguments or on the `this` object. 
+
+```c++
+struct SWIFT_NONESCAPABLE View {
+    View() : member(nullptr) {}
+    View(const int *p [[clang::lifetimebound]]) : member(p) {}
+    View(const View&) = default;
+private:
+    const int *member;
+};
+
+struct Owner {
+    int data;
+
+    View handOutView() const [[clang::lifetimebound]] {
+        return View(&data);
+    }
+};
+
+View getView(const Owner& owner [[clang::lifetimebound]]) {
+    return View(&owner.data);
+}
+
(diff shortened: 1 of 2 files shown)


Xazax-hun:
I cannot add reviewers, so adding some people manually. @egorzhdan @ravikandhadai @j-hui

j-hui (commented):
Some feedback for now; I haven't read the rest of the article yet. But it mostly looks good so far!

documentation/cxx-interop/safe-interop/index.md
```diff
@@ -0,0 +1,297 @@
...
+template<typename T>
+struct SWIFT_ESCAPABLE_IF(T) MyList {
+    ...
+};
+```
+
+## Lifetime annotations in detail
+
+The `lifetimebound` annotation can be used to annotate methods, free functions, and constructors.
+It can annotate dependence on regular arguments or on the `this` object.
```
j-hui, line 201: Is it worth mentioning here that `lifetimebound` is not a Swift-specific thing, but a Clang attribute?

Also nit:

```suggestion
The `lifetimebound` attribute can be used to annotate methods, free functions, and constructors.
It can annotate dependence on regular arguments or on the `this` object:
```

documentation/cxx-interop/safe-interop/index.md
```diff
@@ -0,0 +1,297 @@
...
+    return View(&owner.data);
+}
+
+View getViewFromEither(View view1 [[clang::lifetimebound]], View view2 [[clang::lifetimebound]]) {
+    if (coinFlip)
+        return view1;
+    else
+        return view2;
+}
+```
```
j-hui, line 234: This code block is pretty big, and features several somewhat unrelated examples. It's also not immediately clear to those not already familiar with `lifetimebound` whose lifetime is dependent on whom in each case.

Perhaps you can break this up into several code blocks with some prose in between to describe each?

documentation/cxx-interop/safe-interop/index.md
```diff
@@ -0,0 +1,297 @@
...
+These types might point to static data or might represent an empty sequence or lack of data.
+Such functions need to be annotated with `SWIFT_RETURNS_INDEPENDENT_VALUE`:
+
+```c++
+View returnsEmpty() SWIFT_RETURNS_INDEPENDENT_VALUE {
+    return View();
+}
+```
+
+The default constructor of a type is assumed to create an independent value.
```
j-hui, line 246: This is kind of an orphaned remark.

Maybe:

```suggestion
Notably, the default constructor of a type is always assumed to create an independent value.
```

(Unless annotated otherwise?)

documentation/cxx-interop/safe-interop/index.md
```diff
@@ -0,0 +1,297 @@
...
+        return view2;
+}
+```
+
+Occasionally, a function might return a non-escapable type that in fact has no dependency on any other values.
+These types might point to static data or might represent an empty sequence or lack of data.
+Such functions need to be annotated with `SWIFT_RETURNS_INDEPENDENT_VALUE`:
+
+```c++
+View returnsEmpty() SWIFT_RETURNS_INDEPENDENT_VALUE {
```
j-hui, line 241: Pretending I'm a reader that knows nothing about Swift vs Clang annotations, I might be confused as to why this annotation looks so different from `[[clang::lifetimebound]]`. Perhaps it's worth mentioning that this annotation only means something to imported APIs, and doesn't do anything in pure C++ code? (Unless it does?)

documentation/cxx-interop/safe-interop/index.md
```diff
@@ -0,0 +1,297 @@
...
+    return View();
+}
+```
+
+The default constructor of a type is assumed to create an independent value.
+
+We can annotate `lifetimebound` APIs via APINotes. The -1 index represents the this position.
```
j-hui, line 248:
```suggestion
We can also annotate `lifetimebound` APIs via APINotes. The `-1` index represents the this position.
```

documentation/cxx-interop/safe-interop/index.md
```diff
@@ -0,0 +1,297 @@
...
+All of the non-escapable inputs need lifetime annotations for a function to be
+considered safe. If an input never escapes from the called function we can use
+the `noescape` annotation.
+
+```c++
+void is_palindrome(std::span<int> s [[clang::noescape]]);
+```
+
+## Convenience overloads for annotated spans and pointers
+
```
j-hui, line 297: Is there more coming soon? 🙂

documentation/cxx-interop/safe-interop/index.md
```diff
@@ -0,0 +1,297 @@
...
+```c++
+View returnsEmpty() SWIFT_RETURNS_INDEPENDENT_VALUE {
+    return View();
+}
+```
+
+The default constructor of a type is assumed to create an independent value.
+
+We can annotate `lifetimebound` APIs via APINotes. The -1 index represents the this position.
+Note that APINotes have some limitations around C++, they do not support overloaded functions.
```
j-hui, line 249: Suggestion: move this caveat to after the APINotes example.

documentation/cxx-interop/safe-interop/index.md
```diff
@@ -0,0 +1,297 @@
...
+      Parameters:
+        - Position:      -1
+          Lifetimebound: true
+    - Name: methodToAnnotate
+      Parameters:
+        - Position:      0
+          Lifetimebound: true
+```
+
+We can use `lifetime_capture_by` annotations for output arguments.
```
j-hui, line 265: What does this do? It seems like this section could be fleshed out quite a bit more (or even given its own sub-heading).

documentation/cxx-interop/safe-interop/index.md
```diff
@@ -0,0 +1,297 @@
...
+};
+```
+
+All of the non-escapable inputs need lifetime annotations for a function to be
+considered safe. If an input never escapes from the called function we can use
+the `noescape` annotation.
+
+```c++
+void is_palindrome(std::span<int> s [[clang::noescape]]);
+```
```
j-hui, line 294: Same remark about this section being fleshed more + given its own sub-heading

j-hui (commented):
Completed my first round of comments. Looks good so far, but it seems like the end could be fleshed out a little more!

A broader question: for decls that are separately declared and defined, where should the annotation go? I assume that for types/classes, the annotation should be given at the definition site, while for functions, the annotation should be given at the declaration site. But I'm not sure, and it'd be nice to see this explicitly specified in case a reader would like to organize their code different than in the given examples.

documentation/cxx-interop/safe-interop/index.md
```diff
@@ -0,0 +1,297 @@
...
+    return fileName(normalizedPath)
+}
+```
+
+Building this code will emit a warning that the `fileName` call is unsafe because
+it references the unsafe type `StringRef`. Swift considers `StringRef` unsafe because
+it has a pointer member. Types like `StringRef` can dangle, we need to take extra
+care using them, making sure the referenced buffer outlives the `StringRef` object.
+
+Swift's [non-escapable types](https://github.com/swiftlang/swift-evolution/blob/main/proposals/0446-non-escapable.md)
```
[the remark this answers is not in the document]
Xazax-hun: Unfortunately, I am not aware of any documentation yet :(

documentation/cxx-interop/safe-interop/index.md
```diff
@@ -0,0 +1,297 @@
...
+annotation.
+
+```c++
+struct SWIFT_ESCAPABLE Owner {
+    ...
+}; 
+```
+
+The main reason for explicitly annotating a type as `SWIFT_ESCAPABLE` to make sure
+it is considered as a safe type when used from Swift. Functions returning escapable
```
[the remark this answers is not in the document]
Xazax-hun: No, the Swift compiler would not complain about pointer or other unsafe members. It blindly believes this annotation. I am actually not sure about the non-escapable members! I will need to write a test. Escapable type with a non-escapable member is not valid Swift. I'll test this out and check if we have a missing diagnostic.

documentation/cxx-interop/safe-interop/index.md
Xazax-hun: Yup, I plan to update this PR soon with the rest but wanted to get some early feedback for the first half.

documentation/cxx-interop/safe-interop/index.md
```diff
@@ -0,0 +1,389 @@
...
+{:.no_toc}
+
+* TOC
+{:toc}
+
+## Introduction
+
+Swift's [strict memory safety mode](https://github.com/swiftlang/swift-evolution/blob/main/proposals/0458-strict-memory-safety.md)
+is a new feature in Swift 6.2 to make code easier to audit for memory safety.
+This document describes how to ergonomically interact with Swift by importing
```
ravikandhadai, line 19: I think it might be better to swap the second and first sentences and highlight at the beginning that this in general helps import C++ APIs safely.

documentation/cxx-interop/safe-interop/index.md
```diff
@@ -0,0 +1,389 @@
...
+Future changes to the design or functionality of C++ interoperability will not
+break code in existing codebases [by default](#source-stability-guarantees-for-mixed-language-codebases).
+</div>
+
+## Overview
+
+Swift provides memory safety with a combination of language affordances and runtime checking.
+However, Swift also deliberately includes some unsafe constructs, such as the `UnsafePointer` and `UnsafeMutablePointer`
+types in the standard library.
+Swift occasionally needs additional information that is not present in the C++ type and API declarations
```
ravikandhadai, line 43: I would prefer "in some cases" over "occasionally", as that may imply that this is very infrequent.

documentation/cxx-interop/safe-interop/index.md
```diff
@@ -0,0 +1,389 @@
...
+
+```swift
+func getFileName(_ path: borrowing std.string) -> StringRef {
+    let normalizedPath = normalize(path)
+    return fileName(normalizedPath)
+}
+```
+
+Building this code will emit a warning that the `fileName` call is unsafe because
+it references the unsafe type `StringRef`. Swift considers `StringRef` unsafe because
```
ravikandhadai, line 79: > Swift considers `StringRef` unsafe because it has a pointer member.

Is the definition of unsafe we use in strictly-safe mode same as this definition of a view type [here](https://www.swift.org/documentation/cxx-interop/#working-with-c-references-and-view-types-in-swift):  (C++ Types Considered to Be References or View Types by Swift)?

documentation/cxx-interop/safe-interop/index.md
```diff
@@ -0,0 +1,389 @@
...
+However, Swift also deliberately includes some unsafe constructs, such as the `UnsafePointer` and `UnsafeMutablePointer`
+types in the standard library.
+Swift occasionally needs additional information that is not present in the C++ type and API declarations
+to safely interface with them. This document describes how such code needs to be annotated.
+
+### Annotating foreign types
+
+Types imported from C++ are considered foreign to Swift. Many of these types are considered safe,
+including the built-in integral types like `int`, some standard library types like `std::string`,
+and aggregate types built from other safe types.
```
ravikandhadai, line 50: Does this list cover all safe types? Is this missing out on anything. If it covers all safe types we could say "namely" instead of "including", to clarify this is the definition of what is safe.

documentation/cxx-interop/safe-interop/index.md
```diff
@@ -0,0 +1,389 @@
...
+
+On the other hand, some C++ types are imported as unsafe by default. Consider the following C++ type
```
ravikandhadai, line 52: Is everything that is not safe, unsafe? If so, we can all other C++ types are unsafe by default.

documentation/cxx-interop/safe-interop/index.md
```diff
@@ -0,0 +1,389 @@
...
+
+Swift provides memory safety with a combination of language affordances and runtime checking.
+However, Swift also deliberately includes some unsafe constructs, such as the `UnsafePointer` and `UnsafeMutablePointer`
+types in the standard library.
+Swift occasionally needs additional information that is not present in the C++ type and API declarations
+to safely interface with them. This document describes how such code needs to be annotated.
+
+### Annotating foreign types
+
+Types imported from C++ are considered foreign to Swift. Many of these types are considered safe,
```
ravikandhadai, line 48: >Many of these types are considered safe

As I read through the section on SWIFT_ESCAPABLE annotation, I realized that there is difference between what is safe and unsafe under strict-memory-safety mode and without it. I think it is worth specifically talking about this aspect here e.g.

Types imported from C++ are considered foreign to Swift. Normally, most C++ types are imported into Swift without any restriction to reduce the effort need to call into C++ code. A small set of C++ APIs e.g. pointers/references and methods returning pointers are imported as unsafe (section [Working with C++ references and view types in Swift](url) explains this in more detail.) Under the strict memory safe mode, the compiler will flip the polarity and treat all types that are not known to be safe as unsafe, and will diagnose uses of them. In this section, we will show how to annotate unsafe C++ types so that they can be accessed safely and correctly from Swift. Note that the features here are agnostic to whether strictly-safe mode is on or off.

ravikandhadai: The rest of the paragraph is specific to strict memory safety mode. So, we can start that with, Under the strict memory safety mode, integral types, std library types and aggregated types built from ... Where other types are considered unsafe ...

documentation/cxx-interop/safe-interop/index.md
```diff
@@ -0,0 +1,389 @@
...
+}
+```
+
+After annotating the C++ code, the Swift compiler can enforce the lifetime
+contracts helping us to write code that is free of memory safety errors.
+
+## Escapability annotations in detail
+
+Currently, unannotated types are imported as `Escapable` to maintain backward
+compatibility. This might change in the future under a new interoperability version.
```
ravikandhadai, line 147: >Currently, unannotated types are imported as `Escapable` to maintain backward compatibility.

Is it under strict memory safety mode or without it? I think we are talking about the normal case here. I think we can drop this sentence altogether.

documentation/cxx-interop/safe-interop/index.md
```diff
@@ -0,0 +1,389 @@
...
+
+Moreover, we can explicitly mark types as `Escapable` using the `SWIFT_ESCAPABLE`
+annotation:
+
+```c++
+struct SWIFT_ESCAPABLE Owner { ... }; 
+```
+
+The main reason for explicitly annotating a type as `SWIFT_ESCAPABLE` is to make sure
+it is considered as a safe type when used from Swift. Functions returning escapable
```
ravikandhadai, line 167: Can we add sentence here to say what it means for type to be "escapable" i.e., it is not lifetime dependent on any other type.

documentation/cxx-interop/index.md
```diff
@@ -1331,6 +1331,12 @@ automatically.
 
 ## Working with C++ References and View Types in Swift
 
+Swift has a new [safe interoperability](https://www.swift.org/documentation/cxx-interop/safe-interop)
```
rapidsna, line 1334: @egorzhdan Wouldn't this change expose the link to safe-interop to public?

documentation/cxx-interop/safe-interop/index.md
```diff
@@ -0,0 +1,401 @@
...
+| `void takeSpan(IntSpan x [[clang::noescape]]);`           | `func takeSpan(_ x: Span<Int32>)`                                    |
+| `IntSpan changeSpan(IntSpan x [[clang::lifetimebound]]);` | `@lifetime(x) func changeSpan(_ x: Span<Int32>) -> Span<Int32>`      |
+| `IntSpan changeSpan(IntVec& x [[clang::lifetimebound]]);` | `@lifetime(x) func changeSpan(_ x: borrowing IntVec) -> Span<Int32>` |
+| `IntSpan Owner::getSpan() [[clang::lifetimebound]];`      | `@lifetime(self) func getSpan() -> Span<Int32>`                      |
+
+These transformations only support top level `std::span`s, we do not
+transform the nested cases. A `std::span` of a non-const type `T` will
+be transformed to `MutableSpan<T>` on the Swift wide.
+
+### Annotated pointers
```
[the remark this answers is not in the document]
Xazax-hun: This section is closely related to the bounds safety annotations so I believe that this should be authored by the team that worked on this.

Xazax-hun:
Closing in favour of https://github.com/swiftlang/swift-org-website/pull/980

PR #967, opened 2025-04-24
[SSWG] Add Josh as SSWG member

sebsto:
### Motivation:

Add Josh Elkins in the list of SSWG members

### Modifications:

Modified the `_data/server-workgroup/members.yml` 

### Result:

This page should be updated 
https://www.swift.org/sswg/

### DIFF
--- a/_data/server-workgroup/members.yml
+++ b/_data/server-workgroup/members.yml
@@ -45,3 +45,7 @@
 - name: Ben Cohen
   github: airspeedswift
   affiliation: Apple, Core team rep
+
+- name: Josh Elkins
+  github: jbelkins
+  affiliation: Amazon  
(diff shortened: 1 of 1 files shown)


FranzBusch (commented):
Can you do me a favor and re-order the list to be alphabetical. That's something we had at some point. We should also update the affiliation of Simon while we are at it.

alexandersandberg:
> Can you do me a favor and re-order the list to be alphabetical. That's something we had at some point. We should also update the affiliation of Simon while we are at it.

The code rendering the list already sorts it alphabetically, so the data does not necessarily need to be sorted as well. 🙂 

CleanShot 2025-04-24 at 10 10 40

sebsto:
@alexandersandberg you beat me to it. I updated the list anyway for easier maintenance in the future

0xTim (approved):

PR #980, opened 2025-05-13
[cxx-interop] Add initial documentation for strict memory safety

egorzhdan:
99% of this document is taken from Gabor's PR: https://github.com/swiftlang/swift-org-website/pull/958.

### DIFF
--- a/documentation/cxx-interop/safe-interop/index.md
+++ b/documentation/cxx-interop/safe-interop/index.md
@@ -1,0 +1,412 @@
+---
+layout: page
+title: Safely Mixing Swift and C++
+official_url: https://swift.org/documentation/cxx-interop/safe-interop/
+redirect_from:
+- /documentation/cxx-interop/safe-interop.html
+---
+
+## Table of Contents
+{:.no_toc}
+
+* TOC
+{:toc}
+
+## Introduction
+
+This document describes the additional memory safety features that Swift
+provides when interoperating with C and C++. It describes the ways to make Swift
+aware of certain memory safety aspects of your codebase, as well as how Swift
+uses that knowledge to prevent common memory safety issues in your code.
+
+* * *
+
+<div class="info" markdown="1">
+C++ interoperability is an actively evolving feature of Swift.
+Future releases of Swift might change how Swift and C++
+interoperate,
+as the Swift community gathers feedback from real world adoption of C++
+interoperability in mixed Swift and C++ codebases.
+Please provide the feedback that you have on the
+[Swift Forums](https://forums.swift.org/c/development/c-interoperability/), or
+by filing an [issue on GitHub](https://github.com/swiftlang/swift/issues/new/choose).
+Future changes to the design or functionality of C++ interoperability will not
+break code in existing codebases [by default](#source-stability-guarantees-for-mixed-language-codebases).
+</div>
+
+## Overview
+
+Swift provides memory safety with a combination of language affordances and runtime checking.
+However, Swift also deliberately includes some unsafe constructs, such as the `UnsafePointer` and `UnsafeMutablePointer`
+types in the standard library.
+In some cases, Swift needs additional information that is not present in the C++ type and API declarations
+to safely interface with them. This document describes how such code needs to be annotated so that it can be used in strict safety mode.
+
+### Annotating Foreign Types
+
+Normally, C++ types imported into Swift are assumed to not impact Swift's memory
+safety, and can be used without safety-related restrictions. However, a small
+set of C++ APIs e.g. pointers/references and methods returning pointers will be
+imported as unsafe (section [Working with C++ references and view types in Swift](https://www.swift.org/documentation/cxx-interop/#working-with-c-references-and-view-types-in-swift)
+explains this in more detail.) Under the strict memory safety mode, the compiler will flip the polarity of its safety assumptions.
+It will treat all types that are not known to be safe as unsafe, and emit diagnostics for usage of unsafe types without the `unsafe` keyword.
+In this section,
+we will show how to annotate unsafe C++ types so that they can be accessed safely and correctly from Swift.
+Note that the features here are agnostic to whether strict safety mode is on or off. When the strict safety
+mode is on, the compiler diagnostics can serve as a guide on how to properly annotate C++ types, and also help ensure
+that the code doesn't use unsafe APIs anywhere. When the strict safety mode is turned off, it is still
+recommended to adopt these annotation wherever appropriate, especially on C++ types that are potentially
+dependent on the lifetimes of other objects.
+
+Under the strict safety mode, built-in numeric types like `int`, some standard library types like `std::string`,
+and aggregate types built from other safe types are considered safe. All other unannotated types
+are considered unsafe.
+
+Let's see what happens when we are trying to use an unannotated type in
+strict safety mode. Consider the following C++ type and APIs:
+
+```c++
+struct StringRef {
+public:
+    ...
+private:
+    const char *ptr;
+    size_t len;
+};
+
+std::string normalize(const std::string &path);
+
+StringRef fileName(const std::string &normalizedPath);
+```
+
+Let's try to use them from Swift with strict memory safety enabled:
+
+```swift
+func getFileName(_ path: borrowing std.string) -> StringRef {
+    let normalizedPath = normalize(path)
+    return fileName(normalizedPath)
+}
+```
+
+Building this code will emit a warning that the `fileName` call is unsafe because
+it references the unsafe type `StringRef`. Swift considers `StringRef` unsafe because
+it has a pointer member. Pointers in types like `StringRef` can dangle, so we need to take extra
+care that the buffer they point to outlives the `StringRef` object they are encapsulated by.
+
+Swift's [non-escapable types](https://github.com/swiftlang/swift-evolution/blob/main/proposals/0446-non-escapable.md)
+can have lifetime dependencies. The Swift compiler can track these dependencies
+and enforce safety at compile time. To import `StringRef` as a safe type, we
+need to mark it as a non-escapable type by annotating the class definition:
+
+```c++
+struct SWIFT_NONESCAPABLE StringRef { ... };
+```
+
+Now the Swift compiler imports `StringRef` as a safe type and no longer
+emits a warning about using an unsafe type.
+
+### Annotating C++ APIs
+
+Building the code again will emit a new diagnostic for the `fileName` function about
+missing lifetime annotations. C and C++ functions that return non-escapable types need annotations
+to describe their lifetime contracts via [lifetimebound](https://clang.llvm.org/docs/AttributeReference.html#id8)
+and [lifetime_capture_by](https://clang.llvm.org/docs/AttributeReference.html#lifetime-capture-by) annotations.
+
+```c++
+StringRef fileName(const std::string &normalizedPath [[clang::lifetimebound]]);
+```
+
+Adding this annotation to `fileName` indicates that the returned `StringRef` value has the
+same lifetime as the argument of the `fileName` function.
+
+Building the project again reveals a lifetime error in the Swift function that calls `fileName`:
+
+```swift
+func getFileName(_ path: borrowing std.string) -> StringRef {
+    let normalizedPath = normalize(path)
+    return fileName(normalizedPath)
+    // error: lifetime-dependent value escapes local scope
+    // note: depends on `normalizedPath`
+}
+```
+
+The value returned by `fileName` will dangle after the lifetime of
+`normalizedPath` ends. We can fix this error by making the caller of
+`getFileName` normalize the path prior to calling `getFileName`:
+
+```swift
+// Path needs to be normalized.
+func getFileName(_ normalizedPath: borrowing std.string) -> StringRef {
+    return fileName(normalizedPath)
+}
+```
+
+Alternatively, we could construct and return an `Escapable` value of type such
+as `std.string` instead of a dangling `StringRef`:
+
+```swift
+func getFileName(_ path: borrowing std.string) -> std.string {
+    let normalizedPath = normalize(path)
+    let ref = fileName(normalizedPath)
+    return ref.toString()
+}
+```
+
+After annotating the C++ code, the Swift compiler can enforce those lifetime
+contracts and help us to write code that is free of memory safety errors.
+
+## Escapability Annotations in Detail
+
+Under the strict safety mode, even though compiler warns us when importing
+unannotated types, they are still imported as if they are `Escapable` to
+maintain backward compatibility. This behavior might change in the future under
+a new interoperability version.
+
+We have already seen that we can import a type as `~Escapable` to Swift by adding
+the `SWIFT_NONESCAPABLE` annotation:
+
+```c++
+struct SWIFT_NONESCAPABLE View {
+    View(const int *p) : member(p) {}
+private:
+    const int *member;
+};
+```
+
+Moreover, we can explicitly mark types as `Escapable` using the `SWIFT_ESCAPABLE`
+annotation to express that they are not lifetime dependent on any other values:
+
+```c++
+struct SWIFT_ESCAPABLE Owner { ... }; 
+```
+
+The main reason for explicitly annotating a type as `SWIFT_ESCAPABLE` is to make sure
+it is considered a safe type when used from Swift. Functions returning escapable
+types do not need lifetime annotations.
+
+Escapability annotations can also be attached to types via [API Notes](https://clang.llvm.org/docs/APINotes.html):
+
+```
+Tags:
+- Name: NonEscapableType
+  SwiftEscapable: false
+- Name: EscapableType
+  SwiftEscapable: true
+```
+
+In case of template instantiations the escapability of a type can depend on the
+template arguments:
+
+```c++
+MyList<View> f();
+MyList<Owner> g();
+```
+
+In this example, `MyList<View>` should be imported as `~Escapable` while `MyList<Owner>`
+should be imported as `Escapable`. This can be achieved via conditional escapability
+annotations:
+
+```c++
+template<typename T>
+struct SWIFT_ESCAPABLE_IF(T) MyList {
+    ...
+};
+```
+
+Here, instantiations of `MyList` are imported as `Escapable` when `T` is substituted
+with an `Escapable` type.
+
+The `SWIFT_ESCAPABLE_IF` macro can take multiple template parameters:
+
+```c++
+template<typename F, typename S>
+struct SWIFT_ESCAPABLE_IF(F, S) MyPair {
+    F first;
+    S second;
+};
+```
+
+`MyPair` instantiations are only imported as `Escapable` if both template arguments
+are `Escapable`.
+
+`Escapable` types cannot have `~Escapable` fields. The following code snippet will
+trigger a compiler error:
+
+```c++
+struct SWIFT_NONESCAPABLE View { ... };
+struct SWIFT_ESCAPABLE Owner { 
+    View v;
+}; 
+```
+
+Escapability annotations will not only help the Swift compiler to import C++ types
+safely, it will also help discover missing lifetime annotations as all `~Escapable`
+parameters and return values need to be annotated in an API to make its use safe in
+Swift.
+
+## Lifetime Annotations in Detail
+
+The `lifetimebound` attribute on a function parameter or implicit object parameter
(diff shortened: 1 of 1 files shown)


documentation/cxx-interop/safe-interop/index.md
```diff
@@ -0,0 +1,398 @@
...
+to manually guarantee safety.
+
+## Convenience Overloads for Annotated Spans and Pointers
+
+C++ APIs often using standard library types or other constructs like a
+pointer and a size to represent buffers that have Swift equivalents like
+Swift's `Span` type. These Swift types have additional requirements and
+guarantees. When these properties are properly annotated on the C++ side,
+the Swift compiler can introduce safe convenience functions to make
+interacting with the C++ APIs as effortless as if they were written in Swift.
```
j-hui, line 374:
```suggestion
C++ APIs often feature parameters that denote a span of memory.
For example, some might have two parameters where one points to a memory buffer and the other designates the buffer's size;
others might use the [`std::span`](https://en.cppreference.com/w/cpp/container/span) type from the C++ standard library.
When such APIs are appropriately annotated, the Swift compiler can bridge those span-like parameters to Swift's [`Span`](https://github.com/swiftlang/swift-evolution/blob/main/proposals/0447-span-access-shared-contiguous-storage.md) and [`MutableSpan`](https://github.com/swiftlang/swift-evolution/blob/main/proposals/0467-MutableSpan.md) types, and the user with a safe and convenient interface to those imported APIs.
```

j-hui: We are missing documentation for parameters annotated with `__counted_by` and `__sized_by`.

documentation/cxx-interop/safe-interop/index.md
```diff
@@ -0,0 +1,398 @@
...
+
+### C++ `std::span` Support
+
+APIs taking or returning C++'s `std::span` with sufficient lifetime
+annotations will automatically get overloads taking/returning Swift
+`Span`.
```
j-hui, line 380:
```suggestion
annotations will automatically get safe overloads take or return Swift
`Span` and `MutableSpan` types.
```

documentation/cxx-interop/safe-interop/index.md
```diff
@@ -0,0 +1,398 @@
...
+```
+
+| C++ API                                                   | Generated Swift overload                                             |
+| --------------------------------------------------------- | -------------------------------------------------------------------- |
+| `void takeSpan(IntSpan x [[clang::noescape]]);`           | `func takeSpan(_ x: Span<Int32>)`                                    |
+| `IntSpan changeSpan(IntSpan x [[clang::lifetimebound]]);` | `@lifetime(x) func changeSpan(_ x: Span<Int32>) -> Span<Int32>`      |
+| `IntSpan changeSpan(IntVec& x [[clang::lifetimebound]]);` | `@lifetime(x) func changeSpan(_ x: borrowing IntVec) -> Span<Int32>` |
+| `IntSpan Owner::getSpan() [[clang::lifetimebound]];`      | `@lifetime(self) func getSpan() -> Span<Int32>`                      |
+
+These transformations only support top level `std::span`s. The compiler
```
j-hui, line 396:
```suggestion
These transformations only support top-level `std::span`s. The compiler
```

documentation/cxx-interop/safe-interop/index.md
```diff
@@ -0,0 +1,398 @@
...
+currently does not transform nested `std::span`s. A `std::span` of a non-const
+type `T` is transformed to `MutableSpan<T>` on the Swift wide.
```
j-hui, line 398:
```suggestion
type `T` is transformed to `MutableSpan<T>` on the Swift side.
```

documentation/cxx-interop/safe-interop/index.md
```diff
@@ -0,0 +1,398 @@
...
+
+| C++ API                                                   | Generated Swift overload                                             |
+| --------------------------------------------------------- | -------------------------------------------------------------------- |
+| `void takeSpan(IntSpan x [[clang::noescape]]);`           | `func takeSpan(_ x: Span<Int32>)`                                    |
+| `IntSpan changeSpan(IntSpan x [[clang::lifetimebound]]);` | `@lifetime(x) func changeSpan(_ x: Span<Int32>) -> Span<Int32>`      |
+| `IntSpan changeSpan(IntVec& x [[clang::lifetimebound]]);` | `@lifetime(x) func changeSpan(_ x: borrowing IntVec) -> Span<Int32>` |
+| `IntSpan Owner::getSpan() [[clang::lifetimebound]];`      | `@lifetime(self) func getSpan() -> Span<Int32>`                      |
+
+These transformations only support top level `std::span`s. The compiler
+currently does not transform nested `std::span`s. A `std::span` of a non-const
```
j-hui, line 397:
```suggestion
currently does not transform nested `std::span`s. A `std::span<T>` of a non-const
```

documentation/cxx-interop/safe-interop/index.md
j-hui: If that's only coming in a follow-up patch, then maybe we shouldn't even mention the case where we denote a span with two separate parameters, where one is a pointer and the other is a size/length.

j-hui (changes requested):
I left a bunch of suggested changes.

Some recurring themes:
- We should be consistent about referring to "[the] strict [memory] safety mode" (instead of, e.g., "strict[ly] safe mode.")
- We could be more consistent about code style, e.g., writing `T* x` instead of `T *t`.

There's also no documentation about bridging for `ptrcheck.h` assertions.

Also, I'm wondering how much we should say, "this also works in C by the way." After all, the only C++-specific parts of this document are discussions pertaining to methods, constructors, and `this`, as well the stuff about `std::vector` and `std::span` at the end. I don't think we need to write this to be in C, but perhaps we should at least mention that the parts relevant to C should also work in C (i.e., without C++ interop enabled).

documentation/cxx-interop/safe-interop/index.md
```diff
@@ -0,0 +1,398 @@
...
+definition:
+
+```c++
+class SWIFT_NONESCAPABLE StringRef { ... };
+```
+
+Now the Swift compiler imports `StringRef` as a safe type and no longer
+emits a warning about using an unsafe type.
+
+### Annotating C++ APIs
```
delcypher, line 107: There's no discussion of `__counted_by`, `__counted_by_or_null`, `__sized_by`, or `__sized_by_or_null`, or `__noescape` here. I feel like these should be mentioned.

documentation/cxx-interop/safe-interop/index.md
```diff
@@ -0,0 +1,398 @@
...
+## Convenience Overloads for Annotated Spans and Pointers
+
+C++ APIs often using standard library types or other constructs like a
+pointer and a size to represent buffers that have Swift equivalents like
+Swift's `Span` type. These Swift types have additional requirements and
+guarantees. When these properties are properly annotated on the C++ side,
+the Swift compiler can introduce safe convenience functions to make
+interacting with the C++ APIs as effortless as if they were written in Swift.
+
+### C++ `std::span` Support
```
delcypher, line 376: I don't really know where to fit this in but there are no examples of using `__counted_by` and friends here. I realize they are not as safe as span but we are providing overloads that are safer that their original counter parts.

documentation/cxx-interop/safe-interop/index.md
```diff
@@ -0,0 +1,398 @@
...
+{:toc}
+
+## Introduction
+
+Swift 6.2 introduces [strict memory safety mode](https://github.com/swiftlang/swift-evolution/blob/main/proposals/0458-strict-memory-safety.md),
+an opt-in language feature that prevents common memory safety issues by running
+extra compile-time safety checks and emitting diagnostics.
+
+This document describes the way Swift/C++ interoperability works when strict
+safe mode is enabled in Swift.
```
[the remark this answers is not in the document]
delcypher: Aren't the safe overloads we provide completely separate from the "strict safe mode"? I've been using them without having this mode on. Introducing the interop makes it seems like it's coupled to "strict safe mode" and I don't think that's the case.

delcypher (changes requested):

documentation/cxx-interop/safe-interop/index.md
```diff
@@ -0,0 +1,398 @@
...
+
+## Table of Contents
+{:.no_toc}
+
+* TOC
+{:toc}
+
+## Introduction
+
+Swift 6.2 introduces [strict memory safety mode](https://github.com/swiftlang/swift-evolution/blob/main/proposals/0458-strict-memory-safety.md),
```
delcypher, line 17: Aren't the safe overloads we provide completely separate from the "strict safe mode"? I've been using them without having this mode on. Introducing the interop makes it seems like it's coupled to "strict safe mode" and I don't think that's the case.

delcypher (changes requested):

documentation/cxx-interop/safe-interop/index.md
```diff
@@ -0,0 +1,398 @@
...
+```
+
+Now the Swift compiler imports `StringRef` as a safe type and no longer
+emits a warning about using an unsafe type.
+
+### Annotating C++ APIs
+
+Building the code again will emit a new diagnostic for the `fileName` function about
+missing lifetime annotations. Functions returning non-escapable types need annotations
+to describe their lifetime contracts via [lifetimebound](https://clang.llvm.org/docs/AttributeReference.html#id11)
```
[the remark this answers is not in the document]
egorzhdan: Thanks for noticing this!

documentation/cxx-interop/safe-interop/index.md
```diff
@@ -0,0 +1,398 @@
...
+compatibility. This might change in the future under a new interoperability version.
+We have already seen that we can import a type as `~Escapable` to Swift by adding
+the `SWIFT_NONESCAPABLE` annotation:
+
+```c++
+struct SWIFT_NONESCAPABLE View {
+    View(const int *p) : member(p) {}
+private:
+    const int *member;
+};
```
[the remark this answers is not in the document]
egorzhdan: clang-format and Xcode seem to default to `type *ptr`, but I agree that we should make it consistent across the document.

documentation/cxx-interop/safe-interop/index.md
```diff
@@ -0,0 +1,398 @@
...
+    int data;
+
+    View handOutView() const [[clang::lifetimebound]] {
+        return View(&data);
+    }
+};
+```
+
+Consider a call site like `View v = o.handOutView()`. The `v` object has the same lifetime
+as `o`.
```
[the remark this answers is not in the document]
egorzhdan: Done

documentation/cxx-interop/safe-interop/index.md
```diff
@@ -0,0 +1,398 @@
...
+```
+
+After annotating the C++ code, the Swift compiler can enforce the lifetime
+contracts helping us to write code that is free of memory safety errors.
+
+## Escapability Annotations in Detail
+
+Under the strictly safe mode, even though compiler warns on unannotated types,
+they are imported as if they are `Escapable` to maintain backward
+compatibility. This might change in the future under a new interoperability version.
```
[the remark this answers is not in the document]
egorzhdan: Thanks!

documentation/cxx-interop/safe-interop/index.md
```diff
@@ -0,0 +1,398 @@
...
+
+```c++
+View getOneOfTheViews(const Owner& owner1 [[clang::lifetimebound]], const Owner& owner2
+    View view1 [[clang::lifetimebound]], View view2 [[clang::lifetimebound]]) {
+    if (coinFlip)
+        return View(&owner1.data);
+    if (coinFlip)
+        return view1;
+    else
+        return view2;
```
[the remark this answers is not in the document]
egorzhdan: Yeah, I would expect `coinFlip()` to be non-deterministic based on the name, but I don't feel strongly here.

documentation/cxx-interop/safe-interop/index.md
```diff
@@ -0,0 +1,398 @@
...
+      Parameters:
+        - Position:      -1
+          Lifetimebound: true
+    - Name: methodToAnnotate
+      Parameters:
+        - Position:      0
+          Lifetimebound: true
+```
+
+Note that APINotes have some limitations around C++, they do not support overloaded functions.
```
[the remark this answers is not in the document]
egorzhdan: Agreed on the spacing in `API Notes`, but would it be more natural to say `... notes have ...` rather than `... notes has ...`?

documentation/cxx-interop/safe-interop/index.md
```diff
@@ -0,0 +1,398 @@
...
+However, Swift also deliberately includes some unsafe constructs, such as the `UnsafePointer` and `UnsafeMutablePointer`
+types in the standard library.
+In some cases, Swift needs additional information that is not present in the C++ type and API declarations
+to safely interface with them. This document describes how such code needs to be annotated.
+
+### Annotating Foreign Types
+
+Types imported from C++ are considered foreign to Swift. Normally, most C++ types are imported into Swift
+without any restriction. However, a small set of C++ APIs e.g. pointers/references and methods returning
+pointers will be imported as unsafe (section [Working with C++ references and view types in Swift](https://www.swift.org/documentation/cxx-interop/#working-with-c-references-and-view-types-in-swift)
```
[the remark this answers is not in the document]
j-hui: It should have been rendered as an emdash, i.e., "—", by Markdown... but now I realize that isn't standard Markdown (though it is supported by some extensions).

I've updated the suggestion to just a unicode emdash.

documentation/cxx-interop/safe-interop/index.md
```diff
@@ -0,0 +1,398 @@
...
+Swift provides memory safety with a combination of language affordances and runtime checking.
+However, Swift also deliberately includes some unsafe constructs, such as the `UnsafePointer` and `UnsafeMutablePointer`
+types in the standard library.
+In some cases, Swift needs additional information that is not present in the C++ type and API declarations
+to safely interface with them. This document describes how such code needs to be annotated.
+
+### Annotating Foreign Types
+
+Types imported from C++ are considered foreign to Swift. Normally, most C++ types are imported into Swift
+without any restriction. However, a small set of C++ APIs e.g. pointers/references and methods returning
```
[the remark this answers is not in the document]
egorzhdan: Good point, rephrased it.

documentation/cxx-interop/safe-interop/index.md
```diff
@@ -0,0 +1,398 @@
...
+## Lifetime Annotations in Detail
+
+The `lifetimebound` attribute on a function parameter or implicit object parameter
+indicates that the returned object's lifetime could end when any of the `lifetimebound`
+annotated parameters' lifetime ended.
+This annotation a constructor describes the lifetime of the created object:
+
+```c++
+struct SWIFT_NONESCAPABLE View {
+    View(const int *p [[clang::lifetimebound]]) : member(p) {}
```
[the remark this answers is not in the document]
egorzhdan: Made these consistent across the document 👍

documentation/cxx-interop/safe-interop/index.md
```diff
@@ -0,0 +1,398 @@
...
+private:
+    const int *member;
+};
+```
+
+Moreover, we can explicitly mark types as `Escapable` using the `SWIFT_ESCAPABLE`
+annotation to express that they are not lifetime dependent on any other values:
+
+```c++
+struct SWIFT_ESCAPABLE Owner { ... };
```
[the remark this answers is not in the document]
egorzhdan: Made these consistent across the document 👍

documentation/cxx-interop/safe-interop/index.md
```diff
@@ -0,0 +1,398 @@
...
+    void captureView(View v [[clang::lifetime_capture_by(this)]]) {
+        view = v;
+    }
+
+    void handOut(View &v) const [[clang::lifetime_capture_by(v)]] {
+       v = view; 
+    }
+
+    View view;
+};
```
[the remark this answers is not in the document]
egorzhdan: Done, renamed the field to `containedView`.

documentation/cxx-interop/safe-interop/index.md
```diff
@@ -0,0 +1,413 @@
...
+}
+```
+
+Building this code will emit a warning that the `fileName` call is unsafe because
+it references the unsafe type `StringRef`. Swift considers `StringRef` unsafe because
+it has a pointer member. Pointers in types like `StringRef` can dangle, so we need to take extra
+care that the buffer they point to outlives the `StringRef` object they are encapsulated by.
+
+Swift's [non-escapable types](https://github.com/swiftlang/swift-evolution/blob/main/proposals/0446-non-escapable.md)
+can also have lifetime dependencies, just like `StringRef`. However, the Swift compiler
```
rapidsna: This sounds a bit awkward to me because nothing talked about "StringRef" had a "lifetime dependencies".

egorzhdan: Good point, I rephrased it slightly.

documentation/cxx-interop/safe-interop/index.md
```diff
@@ -0,0 +1,412 @@
...
+func getFileName(_ path: borrowing std.string) -> StringRef {
+    let normalizedPath = normalize(path)
+    return fileName(normalizedPath)
+    // error: lifetime-dependent value escapes local scope
+    // note: depends on `normalizedPath`
+}
+```
+
+The value returned by `fileName` will dangle after the lifetime of `normalizedPath` ends.
+We can fix this error by pushing the task of normalizing a path to the callee:
```
rapidsna: Did you mean should the "caller" of `getFileName` normalize the path in this example?

documentation/cxx-interop/safe-interop/index.md
rapidsna: +1

rapidsna: The content isn't really about strict memory safety mode. I would instead say something like this for the intro:

This document describes safe interoperability between Swift and C/C++ by using annotations to provide missing information in C/C++ types, enabling their safe use.

While optional, the strict memory safe mode introduced in Swift 6.2 can also be used to help find where these annotations should be added.

documentation/cxx-interop/safe-interop/index.md
```diff
@@ -0,0 +1,412 @@
+---
+layout: page
+title: Mixing Strict Memory Safe Swift and C++
```
rapidsna: Suggested title: "Safely mixing C, C++, and Swift" (We need a follow up to add "C" interop part).

egorzhdan: We don't really talk about the C part at this point, so I think we can call this "Safely Mixing Swift and C++". I amended the title.

documentation/cxx-interop/safe-interop/index.md
egorzhdan: Rephrased this, let me know it it's better now

documentation/cxx-interop/safe-interop/index.md
egorzhdan: I rewrote this paragraph to make it a lot more generic.

egorzhdan:
Thanks folks for the feedback. Let's land this now and iterate on the C interop documentation in a separate patch.

egorzhdan:
@swiftlang/website-workgroup can I ask you folks for an approval here?

documentation/cxx-interop/safe-interop/index.md
```diff
@@ -0,0 +1,412 @@
...
+C++ interoperability is an actively evolving feature of Swift.
+Future releases of Swift might change how Swift and C++
+interoperate,
+as the Swift community gathers feedback from real world adoption of C++
+interoperability in mixed Swift and C++ codebases.
+Please provide the feedback that you have on the
+[Swift Forums](https://forums.swift.org/c/development/c-interoperability/), or
+by filing an [issue on GitHub](https://github.com/swiftlang/swift/issues/new/choose).
+Future changes to the design or functionality of C++ interoperability will not
+break code in existing codebases [by default](#source-stability-guarantees-for-mixed-language-codebases).
```
hnrklssn, line 34: Not sure if this will be added later, but I noticed that this link is dangling, just so we don't forget to write this part

egorzhdan:
> How do website visitors find this article? There does not seem to be a link to it from other places on the website.

Yeap, that's right! Since this document talks about a feature that's only available in a development snapshot of the compiler, not a released toolchain, I didn't link to it from the main C++ interop doc page (https://www.swift.org/documentation/cxx-interop/).

PR #1025, opened 2025-06-02
feat(hero-anim): Add mobile hero animation logic. Misc cleanup relate…

jesseaborden:
### Motivation:

On smaller viewports, the hero graphic is cropped offscreen. This could cause a perceived performance issue as the animation playing offscreen is effectively an animation delay. This PR mainly adds logic to move the animation to the visible portion of the graphic. Additionally, this adds some cleanup and comments. 

### Modifications:

- Added animation offset
- Minor CSS cleanup
- Consolidated duplicate animation logic 
- Added explanatory comments

### Result:

Hero animation starts from the visible portion of the container.

### DIFF
--- a/assets/javascripts/new-javascripts/hero.js
+++ b/assets/javascripts/new-javascripts/hero.js
@@ -1,4 +1,4 @@
-const heroAnimation = async () => {
+const heroAnimation = async (animContainer) => {
   const isReduceMotionEnabled = window.matchMedia(
     '(prefers-reduced-motion: reduce)',
   ).matches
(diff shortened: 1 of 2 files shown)


assets/javascripts/new-javascripts/hero.js
```diff
@@ -14,6 +14,11 @@ const heroAnimation = async () => {
     })
   }
 
+  // Skip to visible portion of animation when cropped on small screens
+  const { left, width } = animContainer.getClientRects()[0]
+  const offScreenDelta = Math.abs(left) / width
+  const initProgressVal = 0 + offScreenDelta
```
federicobucchi: Can this be: `const initProgressVal = offScreenDelta`

federicobucchi: Ah you want to make sure it starts from 0

federicobucchi (approved):

assets/javascripts/new-javascripts/hero.js
jesseaborden: Ah, good catch. Fixed that.

jesseaborden: That was what I was thinking, but looking at it more, I don't think it's necessary. I chucked it. Thanks for catching that 🙂

PR #1038, opened 2025-06-04
Hero Animation: Update Ease

jesseaborden:
### Motivation:

The hero animation's ease is a bit too light and causing the start of the animation to look run too fast. This needs to be turned up a bit. 

### Modifications:

- Updated ease in animation script.

### Result:

Hero animation has a stronger ease in / out.


shahmishal:
@jesseaborden can you target main branch?

PR #1045, opened 2025-06-04
Fix code block tabs on mobile

alexandersandberg:
Fixes broken code block tabs.

### Motivation:

The code block tabs don't work properly on mobile:
* On the install page, the Fish tab is not visible
* On the CLI tools page, it doesn't scroll all the way to the last item.

### Modifications:

- Fixed the width of the tabs
  - I'm not sure what the negative 59px was for?

### Install page

|Before|After|
|---|---|
|CleanShot 2025-06-04 at 22 08 14|CleanShot 2025-06-04 at 22 08 22|

### CLI tools page

|Before|After|
|---|---|
|CleanShot 2025-06-04 at 22 08 02|CleanShot 2025-06-04 at 22 07 32|

### DIFF
--- a/assets/stylesheets/new-stylesheets/pages/_get-started.scss
+++ b/assets/stylesheets/new-stylesheets/pages/_get-started.scss
@@ -554,9 +554,10 @@
 
   @media only screen and (max-width: 768px) {
     nav {
+     width: 100%;
+
       ul {
         white-space: nowrap;
-        width: calc(100% - 59px);
       }
     }
   }


assets/stylesheets/new-stylesheets/pages/_get-started.scss
```diff
@@ -554,9 +554,10 @@
 
   @media only screen and (max-width: 768px) {
     nav {
+     width: 100%;
+
       ul {
         white-space: nowrap;
-        width: calc(100% - 59px);
```
jesseaborden, line 559: Hey @alexandersandberg, The intent here was to extend the tab nav to the edge of the container, so it was clear that the element was scrollable. The 59px accounts for half the horizontal padding. It's definitely a bit of a magic number right now, so maybe we can rework that or add a comment to make it more clear. I do see that the last item isn't scrolling correctly. Adding some padding to the last item or edge of the container may correct that.

jesseaborden: Ah, I was too slow.

alexandersandberg: I see! I think it would be good to avoid the magic numbers though as they're likely to break sooner or later. I will see if I can figure out a way around it.

PR #1048, opened 2025-06-04
Remove links to dead embedded content

rauhul:
Embedded guides are no longer hosted directly on the website.


rauhul:
The new embedded home page has all the relevant links now :)

PR #1049, opened 2025-06-04
Update embedded links

rauhul:
Updates links from apple/swift-foo-examples to
swiftlang/swift-foo-examples for embedded repos.


getting-started/embedded-swift/index.md
```diff
@@ -1,223 +0,0 @@
----
-layout: page
-title: Build an embedded application for a microcontroller
----
```
shahmishal: Can we add a redirect to new embedded use case page?

shahmishal: for example:
```
---
redirect_to: /get-started/embedded/
---
```

PR #1056, opened 2025-06-05
Improve title format for better SEO and readability

alemohamad:
### Motivation:

The homepage currently shows `Swift.org - Welcome to Swift.org` as the `<title>`, which is redundant and not ideal for SEO. Internal pages use the format `Swift.org - Page Title`, placing the main topic last, which is suboptimal for search engine visibility and user experience.

swift-org-seo

Additionally, using multiple hyphens in titles such as `Swift.org - Install Swift - macOS` can make it harder to visually parse where the subject ends and the site name begins. Replacing the separator with an em dash (`—`) improves clarity by visually distinguishing the content title from the site name.

### Result:

This PR proposes changing the homepage title to `Swift Programming Language`, as it offers a more descriptive label for search engines and link previews. However, this is open to discussion — alternatives like `Welcome to Swift.org` may also be appropriate, depending on how the community prefers to present the project. Personally, I believe the proposed title better highlights the language itself and improves discoverability, but it's not intended as a final decision.

Internal page titles now place the page subject first, followed by the site name separated by an em dash, improving structure, clarity, and SEO performance.

swift-org-new-titles

### DIFF
--- a/_layouts/base.html
+++ b/_layouts/base.html
@@ -3,7 +3,7 @@
 
 <head>
   <meta charset="utf-8" />
-  <title>{{ page.title | escape }} — {{ site.title }}</title>
+  <title>{{ page.title | escape }} | {{ site.title }}</title>
   {% if jekyll.environment == 'stage' %}
   <meta name="robots" content="noindex" />
   {% endif %}
--- a/_layouts/new-layouts/base.html
+++ b/_layouts/new-layouts/base.html
@@ -4,7 +4,7 @@
     <meta charset="utf-8" />
     <title>
       {% if page.url == "/" %} {{ page.title | escape }} {% else %} {{
-      page.title | escape }} — {{ site.title }} {% endif %}
+      page.title | escape }} | {{ site.title }} {% endif %}
     </title>
     {% if jekyll.environment == 'stage' %}
     <meta name="robots" content="noindex" />


alemohamad:
@davelester Thanks for your feedback. Yes, I also thought to use a pipe (|) as a separator, but I thought another symbol to not confuse with the letter I. But it's a separator nonetheless. In my opinion we shouldn't have Swift.org at the homepage title, but if the team thinks it adds value to have it, I can agree on that. The main goal was to not repeat Swift.org in the homepage. 😉

_layouts/new-layouts/base.html
```diff
@@ -2,7 +2,10 @@
 <html lang="en">
   <head>
     <meta charset="utf-8" />
-    <title>{{ site.title }} - {{ page.title | escape }}</title>
+    <title>
+      {% if page.url == "/" %} {{ page.title | escape }} {% else %} {{
+      page.title | escape }} — {{ site.title }} {% endif %}
```
davelester: Could this be updated to use a pipe instead of an em dash?

_layouts/base.html
```diff
@@ -3,7 +3,7 @@
 
 <head>
   <meta charset="utf-8" />
-  <title>{{ site.title }} - {{ page.title | escape }}</title>
+  <title>{{ page.title | escape }} — {{ site.title }}</title>
```
davelester: Could this be updated to use a pipe instead of an em dash?

_layouts/new-layouts/base.html
alemohamad: Updated.

_layouts/base.html
alemohamad: Updated.

alemohamad:
Of course, @davelester! That feels right. Let's see which proposals come up during the next meeting.

davelester (approved):

PR #1058, opened 2025-06-05
Remove stray apostrophe in CLI Tools page

natecook1000:
The CLI usage tab seems like it includes an extra apostrophe at the end.


alexandersandberg (approved):
👍

PR #1070, opened 2025-06-07
Add h2 anchors where missing

federicobucchi:
### Motivation:

We have `## title` that are automatically creating anchors for deep links.
Some are not because they are in HTML files.

### Modifications:

Added script to get the `h2` with a specific class to have the anchor

### Result:

Added class for `h2` I thought should be with anchor (not all).
We can do also for others later on, if we think are helpful.
We can also think to do it for some `h3`.


kaishin (approved):

shahmishal:
@federicobucchi can you resolve the conflicts?

PR #1071, opened 2025-06-07
Safe c interop

hnrklssn:
This adds documentation around __counted_by, __sized_by, API notes, and using lifetime attributes in C.
This is stacked on top of https://github.com/swiftlang/swift-org-website/pull/980.

### DIFF
--- a/documentation/cxx-interop/safe-interop/index.md
+++ b/documentation/cxx-interop/safe-interop/index.md
@@ -1,6 +1,6 @@
 ---
 layout: page
-title: Safely Mixing Swift and C++
+title: Safely Mixing Swift and C/C++
 official_url: https://swift.org/documentation/cxx-interop/safe-interop/
 redirect_from:
 - /documentation/cxx-interop/safe-interop.html
@@ -429,7 +429,7 @@ The most common bounds attribute is `__counted_by`. You can apply `__counted_by`
 and return values to indicate the number of elements that the pointer points to, like this:
 
 ```c
-int calculate_sum(const int * __counted_by(len) values, int len);
+int calculate_sum(const int * __counted_by(len) values [[clang::noescape]], int len);
 ```
 
 In this example the function signature on the C side hasn't changed, but the `__counted_by(len)`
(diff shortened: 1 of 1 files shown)


rapidsna:
CC @egorzhdan @Xazax-hun @DougGregor

rauhul:
Have you considered putting this documentation in a docc catalog?

hnrklssn:
> Have you considered putting this documentation in a docc catalog?

I have not! I can't speak for Gábor and Egor, but I simply kept writing in the established file. What are the benefits of a docc catalog? I haven't contributed to the website before, so I don't really know how things work

documentation/cxx-interop/safe-interop/index.md
```diff
@@ -0,0 +1,541 @@
...
+
+| C API                                                                                                 | Generated Swift overload                                             |
+| ----------------------------------------------------------------------------------------------------- | -------------------------------------------------------------------- |
+| `void take_ptr(const int * __counted_by(len) x __noescape, int len);`                                 | `func take_ptr(_ x: Span<Int32>)`                                    |
+| `const int * __counted_by(len) change_ptr(const int * __counted_by(len) x __lifetimebound, int len);` | `@lifetime(x) func change_ptr(_ x: Span<Int32>) -> Span<Int32>`      |
+
+These overloads provide the same bounds safety as their `UnsafeBufferPointer` equvalents, but with
+added lifetime safety. If lifetime information is available the generated safe overload will always
+choose to use `Span` - no `UnsafeBufferPointer` overload will be generated in this case. This means
+that existing callers are not affected by annotating an API with `__counted_by`, but callers using the
```
Xazax-hun, line 574: Is there upstream documentation for these annotations in Clang? Do we want to link to that?

hnrklssn: There are upstream docs for `__counted_by`, but they are outdated, since only a limited set of use cases are currently supported in upstream. The upstream docs currently only list support for flexible array members, which we don't import to Swift in the first place.

rapidsna: You can refer to this documentation https://clang.llvm.org/docs/BoundsSafety.html

documentation/cxx-interop/safe-interop/index.md
```diff
@@ -410,3 +415,128 @@ using IntVec = std::vector<int>;
...
+
+### Safe Overloads for Pointers
+
+If an API uses raw pointers rather than `std::span` - perhaps because it's written in C or Objective-C,
+or because it's an older C++ API that doesn't want to break backwards compatibility -
+it can still receive the same interop safety as `std::span`. This added bounds safety doesn't break
+source compatiblity, nor does it affect ABI. Instead it leverages bounds attributes to express the
+pointer bounds in terms of other parameters in the function signature.
+
+#### Annotating Pointers with Bounds Attributes
```
rapidsna: This elaborates "UnsafeBufferPointer" in a great length, which is not the most recommended usage. I'd elaborate all these with "Span". Since lifetime annotations are already introduced for C++ span, you can just mention that it's still necessary without still focusing on the bounds annotations.

And then "UnsafeBufferPointer" can be briefly mentioned afterwards like -- if lifetime annotations are missing, it's imported as "UnsafeBufferPointer" instead of "Span".

documentation/cxx-interop/safe-interop/index.md
```diff
@@ -410,3 +414,136 @@ using IntVec = std::vector<int>;
...
+
+The most common bounds attribute is `__counted_by`. You can apply `__counted_by` to pointer parameters
+and return values to indicate the number of elements that the pointer points to, like this:
+
+```c
+int calculate_sum(const int * __counted_by(len) values [[clang::noescape]], int len);
```
Xazax-hun: I wonder if the header that needs to be included should be mentioned upfront instead of at the very end.

documentation/cxx-interop/safe-interop/index.md
```diff
@@ -410,3 +414,136 @@ using IntVec = std::vector<int>;
...
+```
+
+In this example the function signature on the C side hasn't changed, but the `__counted_by(len)`
```
Xazax-hun:
```suggestion
In this example, the function signature on the C side hasn't changed, but the `__counted_by(len)`
```

documentation/cxx-interop/safe-interop/index.md
```diff
@@ -410,3 +414,136 @@ using IntVec = std::vector<int>;
...
+                 int len);
+```
+This safe overload will trap if `a.count != b.count || b.count != out.count`:
+```swift
+func sum_vectors(_ a: Span<CInt>,
+                 _ b: Span<CInt>,
+                 _ out: MutableSpan<CInt>)
+```
+
+If one of the `Span` parameters is larger than the others and you intentionally want to use
```
Xazax-hun: This paragraph is more about how to use the span which might be out of scope for us. I wonder if we should just link to Span's documentation instead (if it already exists).

documentation/cxx-interop/safe-interop/index.md
Xazax-hun: Should we mention early on that we only get the `Span` overload for non-escaping parameters?

documentation/cxx-interop/safe-interop/index.md
```diff
@@ -410,3 +414,136 @@ using IntVec = std::vector<int>;
...
+  - Name:              bar
+    BoundsSafety:
+        Kind:          sized_by
+        BoundedBy:     "size"
+```
+
+#### Limitations
+Bounds attributes are not supported for nested pointers: only the outermost pointer can be transformed.
+
+`lifetime_capture_by` is currently not taken into account when generating safe overloads.
```
Xazax-hun: We should provide a link to the documentation of this attribute.

Xazax-hun (approved):
Looks good to me, some nits inline.

Xazax-hun:
I don't have a strong opinion on the formatting, looks like all approaches have pros and cons :/

documentation/cxx-interop/safe-interop/index.md
rapidsna: I suggest we use `__noescape` and `__lifetimebound` which was how we present it in the WWDC talk, instead of the C++11 and C23 style attribute `[[clang::noescape]]` for consistency throughout the whole document.

rapidsna: `__noescape` is also defined in `lifetimebound.h`.

documentation/cxx-interop/safe-interop/index.md
```diff
@@ -410,3 +414,136 @@ using IntVec = std::vector<int>;
 These transformations only support top-level `std::span`s. The compiler
 currently does not transform nested `std::span`s. A `std::span<T>` of a non-const
 type `T` is transformed to `MutableSpan<T>` on the Swift side.
+
+### Safe Overloads for Pointers
+
+If an API uses raw pointers rather than `std::span` - perhaps because it's written in C or Objective-C,
+or because it's an older C++ API that doesn't want to break backwards compatibility -
+it can still receive the same interop safety as `std::span`. This added bounds safety doesn't break
+source compatiblity, nor does it affect ABI. Instead it leverages bounds attributes to express the
```
rapidsna: Nit: The existing text added by @Xazax-hun uses "annotations", but the new texts say "attributes". Should we make it consistent?

documentation/cxx-interop/safe-interop/index.md
rapidsna: I think this is still relevant because this is about the `count` integer parameter that safe overloads collapse into the part of the `Span` parameter.

Though maybe we should make it more clear that "the count of the whole `Span` isn't always the same as the count that the C function expects".

rapidsna:
+1 for the 3rd option, despite horizontal scrolling.

PR #1076, opened 2025-06-09
Fix broken links

federicobucchi:
### Motivation:

Some links give 404

### Result:

Links are fixed


alexandersandberg (commented):
What's the fix here? It looks like a variable is just being renamed?

The links seem to work for me already as well. Where exactly are they broken?

federicobucchi:
> What's the fix here? It looks like a variable is just being renamed?
> 
> The links seem to work for me already as well. Where exactly are they broken?

If you try `Development Snapshots` in https://www.swift.org/install/linux/debian/12/ clicking on `Download (x86_64)` it gives 404.  In the url there is `-aarch64`.

This is happening because the import of `{% include install/_older-releases.md platform=include.platform %}` uses the same variable name that happends `-aarch64`.

Changing the name, we avoid the conflict

alexandersandberg (approved):
Ah, I see! Appreciate the additional detail. Fix looks good 👍

PR #1077, opened 2025-06-10
[cxx-interop] Add SWIFT_RETUNRS_(UN)RETAINED discussions to C++ interop docs

fahadnayyar:
Documenting the usage of `SWIFT_RETUNRS_(UN)RETAINED` annotations to specify the ownership convention of the returned `SWIFT_SHARED_REFERENCE` types from C++ functions and methods

rdar://116817243

### DIFF
--- a/documentation/cxx-interop/index.md
+++ b/documentation/cxx-interop/index.md
@@ -1255,7 +1255,7 @@ object.doSomething()
 // `object` will be released here.
 ```
 
-#### Calling conventions for returning Shared Reference Types
+#### Calling conventions when returning Shared Reference Types from C++ to Swift
 
 When C++ functions and methods return `SWIFT_SHARED_REFERENCE` types, it is necessary to specify the ownership of the returned value.
 For this you should use the `SWIFT_RETURNS_RETAINED` and `SWIFT_RETURNS_UNRETAINED` annotations on functions and methods.
--- a/documentation/cxx-interop/index.md
+++ b/documentation/cxx-interop/index.md
@@ -1261,7 +1261,7 @@ To specify the ownership of returned `SWIFT_SHARED_REFERENCE` types, use the `SW
 // Returns +1 ownership; Swift will take responsibility for releasing it.
 SharedObject* makeOwnedObject() SWIFT_RETURNS_RETAINED;
 
-// Returns +0 ownership; the caller must ensure the object stays alive.
+// Returns +0 ownership; the caller must ensure the object stays alive by retaining it before use.
 SharedObject* makeUnownedObject() SWIFT_RETURNS_UNRETAINED;
 ```
 
@@ -1271,12 +1271,12 @@ let owned = makeOwnedObject()
 owned.doSomething()
 // `owned` is automatically released when it goes out of scope
 
-let unowned = makeUnownedObject()
-unowned.doSomething()
-// make sure `unowned` remains valid while in use
+let unOwned = makeUnownedObject()
+unOwned.doSomething()
+// Swift makes sure `unOwned` remains valid while in use by calling an explicit retain operation.
 ```
 
-These ownership annotations are also supported in Objective-C or Objective-C++ functions that return C++ `SWIFT_SHARED_REFERENCE` types.
+These ownership annotations are also supported in Objective-C++ functions that return C++ `SWIFT_SHARED_REFERENCE` types.
 
 
 ### Inheritance and Virtual Member Functions
--- a/documentation/cxx-interop/index.md
+++ b/documentation/cxx-interop/index.md
@@ -1255,29 +1255,32 @@ object.doSomething()
 // `object` will be released here.
 ```
 
-To specify the ownership of returned `SWIFT_SHARED_REFERENCE` types, use the `SWIFT_RETURNS_RETAINED` and `SWIFT_RETURNS_UNRETAINED` annotations on C++ functions and methods. These annotations tell the Swift compiler whether the type is returned as `+1` (retained) or `+0` (unretained). This is necessary to ensure that appropriate `retain`/`release` operations are inserted at the boundary:
+#### Calling convention for returning shared reference types
+
+When C++ functions and methods return `SWIFT_SHARED_REFERENCE` types, it is necessary to specify the ownership of the returned value. For this you can use the `SWIFT_RETURNS_RETAINED` and `SWIFT_RETURNS_UNRETAINED` annotations on C++ functions and methods. These annotations tell the Swift compiler whether the type is returned as `+1` (retained) or `+0` (unretained). This is necessary to ensure that appropriate `retain`/`release` operations are inserted at the boundary:
 
 ```c++
 // Returns +1 ownership; Swift will take responsibility for releasing it.
 SharedObject* makeOwnedObject() SWIFT_RETURNS_RETAINED;
 
 // Returns +0 ownership; the caller must ensure the object stays alive by retaining it before use.
-SharedObject* makeUnownedObject() SWIFT_RETURNS_UNRETAINED;
+SharedObject* getUnownedObject() SWIFT_RETURNS_UNRETAINED;
 ```
 
 These ownership conventions are reflected naturally in Swift:
 ```swift
 let owned = makeOwnedObject()
 owned.doSomething()
-// `owned` is automatically released when it goes out of scope
+// `owned` is automatically released when it goes out of scope.
 
-let unOwned = makeUnownedObject()
+let unOwned = getUnownedObject()
 unOwned.doSomething()
-// Swift makes sure `unOwned` remains valid while in use by calling an explicit retain operation.
+// Swift makes sure `unOwned` remains valid while in use by calling a retain operation.
 ```
 
 These ownership annotations are also supported in Objective-C++ functions that return C++ `SWIFT_SHARED_REFERENCE` types.
 
+Note that the Swift compiler will automatically infer the ownership conventons for Swift functions returning `SWIFT_SHARED_REFERENCE` types. See [Exposing C++ Shared Reference Types back from Swift](#exposing-c-shared-reference-types-back-from-swift) for calling Swift functions returning `SWIFT_SHARED_REFERENCE` types from C++.
 
 ### Inheritance and Virtual Member Functions
 
--- a/documentation/cxx-interop/index.md
+++ b/documentation/cxx-interop/index.md
@@ -1255,6 +1255,30 @@ object.doSomething()
 // `object` will be released here.
 ```
 
+To specify the ownership of returned `SWIFT_SHARED_REFERENCE` types, use the `SWIFT_RETURNS_RETAINED` and `SWIFT_RETURNS_UNRETAINED` annotations on C++ functions and methods. These annotations tell the Swift compiler whether the type is returned as `+1` (retained) or `+0` (unretained). This is necessary to ensure that appropriate `retain`/`release` operations are inserted at the boundary:
+
+```c++
+// Returns +1 ownership; Swift will take responsibility for releasing it.
+SharedObject* makeOwnedObject() SWIFT_RETURNS_RETAINED;
+
+// Returns +0 ownership; the caller must ensure the object stays alive.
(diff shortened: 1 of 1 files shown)


documentation/cxx-interop/index.md
```diff
@@ -1255,6 +1255,30 @@ object.doSomething()
 // `object` will be released here.
 ```
 
+To specify the ownership of returned `SWIFT_SHARED_REFERENCE` types, use the `SWIFT_RETURNS_RETAINED` and `SWIFT_RETURNS_UNRETAINED` annotations on C++ functions and methods. These annotations tell the Swift compiler whether the type is returned as `+1` (retained) or `+0` (unretained). This is necessary to ensure that appropriate `retain`/`release` operations are inserted at the boundary:
```
Xazax-hun: Would these annotations work on C, and Objective-C code? If yes, maybe we should not emphasize C++ here (also this is the C++ interop documentation so this might be redundant information anyway.)

documentation/cxx-interop/index.md
```diff
@@ -1255,6 +1255,30 @@ object.doSomething()
...
+
+These ownership conventions are reflected naturally in Swift:
+```swift
+let owned = makeOwnedObject()
+owned.doSomething()
+// `owned` is automatically released when it goes out of scope
+
+let unowned = makeUnownedObject()
+unowned.doSomething()
+// make sure `unowned` remains valid while in use
```
Xazax-hun: Wouldn't Swift automatically increase reference counting operations to guarantee that unowned is valid? I think we should check with someone if we are unsure before adding this to the documentation.

documentation/cxx-interop/index.md
```diff
@@ -1255,6 +1255,30 @@ object.doSomething()
...
+```
+
+These ownership annotations are also supported in Objective-C or Objective-C++ functions that return C++ `SWIFT_SHARED_REFERENCE` types.
```
Xazax-hun: Is it possible for an Objective-C function to return a C++ type? Isn't that already Objective-C++?

documentation/cxx-interop/index.md
```diff
@@ -1255,6 +1255,30 @@ object.doSomething()
...
+
+
```
Xazax-hun, line 1286: Is the default ownership convention documented somewhere? If not, this might be a good place to add.

documentation/cxx-interop/index.md
```diff
@@ -1255,6 +1255,30 @@ object.doSomething()
...
+SharedObject* makeUnownedObject() SWIFT_RETURNS_UNRETAINED;
+```
+
+These ownership conventions are reflected naturally in Swift:
+```swift
+let owned = makeOwnedObject()
+owned.doSomething()
+// `owned` is automatically released when it goes out of scope
+
+let unowned = makeUnownedObject()
```
egorzhdan: Minor: looks like the syntax highlighter treats `unowned` as a keyword here, could we pick a different name to avoid the confusion on reader's side?

documentation/cxx-interop/index.md
fahadnayyar: > Would these annotations work on C, and Objective-C code?

The expectation is that they should work, but I'm not aware of any adoption in C and Objective-C code yet. I've only heard adoption feedback from C++ code.

> also this is the C++ interop documentation so this might be redundant information anyway.

I'm not sure I understood this comment. Could you please elaborate on it?

documentation/cxx-interop/index.md
fahadnayyar: Changing `unowned` to `unOwned`

documentation/cxx-interop/index.md
fahadnayyar: Makes sense! Objective-C++ is enough.

documentation/cxx-interop/index.md
fahadnayyar: > // make sure `unowned` remains valid while in use

Oh yes, this comment is misleading.

> Wouldn't Swift automatically increase reference counting operations to guarantee that unowned is valid?

Yes exactly!

Changing the comment to this now:
``// Swift makes sure `unOwned` remains valid while in use by calling an explicit retain operation.``

documentation/cxx-interop/index.md
fahadnayyar: I believe the plan is to make these annotations mandatory in the upcoming release because the default have been a source of confusion and memory safety issues, such as use-after-free memory and memory leaks. Therefore, we should avoid conveying any information about the default in our documentation. Ideally, we would like to warn users not to rely on the default settings.

Ref: https://forums.swift.org/t/stop-importing-unannotated-c-apis-returning-swift-shared-referenece-types-into-swift/74669

fahadnayyar:
CC: @j-hui can you also please take a look?

documentation/cxx-interop/index.md
Xazax-hun: I don't have strong feelings, but I wonder if we should develop a consistent language for this document. E.g., whenever something is C++ specific we say C++, otherwise we say something like foreign function or C family or something else. But if we want to make that change it might be independent of this PR so feel free to leave this paragraph as is.

fahadnayyar:
CC: @ravikandhadai

documentation/cxx-interop/index.md
fahadnayyar: > it might be independent of this PR

Makes sense!
Let's track this problem in a separate ticket: [rdar://154859446](rdar://154859446)


> I don't have strong feelings, but I wonder if we should develop a consistent language for this document. E.g., whenever something is C++ specific we say C++, otherwise we say something like foreign function or C family or something else.

I think we need to have discussion about this offline with more stakeholders. This might be riskier as some things like C foreign reference types exists, but behind an experimental feature flag and have not seem much adoption yet. 

Perhaps we should document C-related information in a separate document. Alternatively, we could revamp this document to include information for both C and C++. However, both of these options necessitate further offline discussions.

CC: @j-hui, @egorzhdan , @ravikandhadai

j-hui: I agree that we should have an offline discussion about how best to organize documentation for C vs C++ interop.

However, independent of that, I think we can simply not mention the language here:

```diff
-use the `SWIFT_RETURNS_RETAINED` and `SWIFT_RETURNS_UNRETAINED` annotations on C++ functions and methods
+use the `SWIFT_RETURNS_RETAINED` and `SWIFT_RETURNS_UNRETAINED` annotations on functions and methods
```

The fact that we are referring to C++ (or C) functions and methods is implicit from the context of this documentation.

Doing so also gives us less to clean up when we _do_ decide how best to approach C vs C++ interop documentation.

documentation/cxx-interop/index.md
```diff
@@ -1255,6 +1255,30 @@ object.doSomething()
...
+
+```c++
+// Returns +1 ownership; Swift will take responsibility for releasing it.
+SharedObject* makeOwnedObject() SWIFT_RETURNS_RETAINED;
+
+// Returns +0 ownership; the caller must ensure the object stays alive by retaining it before use.
+SharedObject* makeUnownedObject() SWIFT_RETURNS_UNRETAINED;
+```
+
+These ownership conventions are reflected naturally in Swift:
```
j-hui: I don't think "naturally" is the right word choice here. "Automatically" might be more appropriate here, but I think it would be even better if you just brought the earlier sentence here:

```suggestion
These annotations are necessary to ensure that appropriate `retain`/`release` operations are inserted at the boundary:
```

documentation/cxx-interop/index.md
```diff
@@ -1255,6 +1255,30 @@ object.doSomething()
...
+```swift
+let owned = makeOwnedObject()
+owned.doSomething()
+// `owned` is automatically released when it goes out of scope
+
+let unOwned = makeUnownedObject()
+unOwned.doSomething()
+// Swift makes sure `unOwned` remains valid while in use by calling an explicit retain operation.
```
j-hui: Two changes I'm suggesting here:
1. `unOwned` -> `unowned` to follow the capitalization of `makeUnownedObject()` (and in any case, "un" and "owned" are not two separate words, so they should not be capitalized separately)
2. I rephrased the comments to be more balanced and only talk about retain operations (so that they parallel one another)


```suggestion
let owned = makeOwnedObject()
// `owned` is already at +1, so no further retain is needed here
owned.doSomething()

let unowned = makeUnownedObject()
// Swift inserts a retain operation on `unowned` here to bring it to +1
unowned.doSomething()
```

documentation/cxx-interop/index.md
```diff
@@ -1255,6 +1255,30 @@ object.doSomething()
...
+```
+
+These ownership annotations are also supported in Objective-C++ functions that return C++ `SWIFT_SHARED_REFERENCE` types.
```
j-hui: I don't think this sentence is actually necessary if we are less specific about what language the annotated functions and methods come from (see my comment further above)

documentation/cxx-interop/index.md
```diff
@@ -1255,6 +1255,30 @@ object.doSomething()
...
 ```
 
+To specify the ownership of returned `SWIFT_SHARED_REFERENCE` types, use the `SWIFT_RETURNS_RETAINED` and `SWIFT_RETURNS_UNRETAINED` annotations on C++ functions and methods. These annotations tell the Swift compiler whether the type is returned as `+1` (retained) or `+0` (unretained). This is necessary to ensure that appropriate `retain`/`release` operations are inserted at the boundary:
+
+```c++
+// Returns +1 ownership; Swift will take responsibility for releasing it.
+SharedObject* makeOwnedObject() SWIFT_RETURNS_RETAINED;
+
+// Returns +0 ownership; the caller must ensure the object stays alive by retaining it before use.
+SharedObject* makeUnownedObject() SWIFT_RETURNS_UNRETAINED;
```
j-hui: I don't think you need to elaborate too much in the comments here:

```suggestion
// Returns +1 ownership
SharedObject* makeOwnedObject() SWIFT_RETURNS_RETAINED;

// Returns +0 ownership
SharedObject* makeUnownedObject() SWIFT_RETURNS_UNRETAINED;
```

In particular, "Swift will take responsibility for releasing it" for the +1 case might be misunderstood to imply that Swift does _not_ take responsibility for releasing the object in the +0 case. Instead, this is true for all FRTs and is not directly relevant to these annotations.

documentation/cxx-interop/index.md
j-hui: I've paraphrased your writing to make sure this says "you should", which is important because in the future we intend to emit warnings when these annotations are absent:

```suggestion
You should annotate functions that return `SWIFT_SHARED_REFERENCE` types
to explicitly specify their return ownership convention.
The `SWIFT_RETURNS_RETAINED` and `SWIFT_RETURNS_UNRETAINED` annotations respectively tell the Swift compiler whether a function returns an object
as `+1` (retained) or `+0` (unretained):
```

j-hui (commented):
Thanks for documenting these annotations! I left some suggestions for how you can paraphrase things to be a little clearer.

documentation/cxx-interop/index.md
fahadnayyar: unOwned -> unowned
It was `unowned` originally which I changed to `unOwned` because `unowned` was appearing as a keyword in the Swift syntatx. This was pointed to me by Egor.

documentation/cxx-interop/index.md
```diff
@@ -1255,6 +1255,35 @@ object.doSomething()
...
+
+let unOwned = getUnOwnedObject()
+// Swift inserts a retain operation on `unowned` here to bring it to +1.
+unOwned.doSomething()
+```
+
+Note that the Swift compiler will automatically infer the ownership conventons for Swift functions returning `SWIFT_SHARED_REFERENCE` types.
+See [Exposing C++ Shared Reference Types back from Swift](#exposing-c-shared-reference-types-back-from-swift) for calling Swift functions returning `SWIFT_SHARED_REFERENCE` types from C++.
+
 ### Inheritance and Virtual Member Functions
```
fahadnayyar, line 1287: TODO: ### -> ####

fahadnayyar:
including this in [PR-1106](https://github.com/swiftlang/swift-org-website/pull/1106)

PR #1078, opened 2025-06-10
[cxx-interop] Add documentation about inferring SWIFT_SHARED_REFERENCE in c++ inheritance

fahadnayyar:
Adding a sentence in the documentation thatthe  foreign-reference annotations are now propagated to inherited C++ types

### DIFF
--- a/documentation/cxx-interop/index.md
+++ b/documentation/cxx-interop/index.md
@@ -1255,6 +1255,9 @@ object.doSomething()
 // `object` will be released here.
 ```
 
+### Inference for Derived Types
+Swift compiler automatically infers the ```SWIFT_SHARED_REFERENCE``` annotation for C++ types that inherit from a base type annotated with ```SWIFT_SHARED_REFERENCE```. These derived types are imported as Swift reference types or Swift class types that use the same ```retain``` and ```release``` functions as the base type. No additional annotation is required on the derived types.
+
 ### Inheritance and Virtual Member Functions
 
 Similar to value types, casting an instance of a derived reference type to a
--- a/documentation/cxx-interop/index.md
+++ b/documentation/cxx-interop/index.md
@@ -1255,9 +1255,10 @@ object.doSomething()
 // `object` will be released here.
 ```
 
-### Inference for Derived Types
+### Inference of Shared Reference behaviour in Derived Types
 
-Swift compiler automatically infers the ```SWIFT_SHARED_REFERENCE``` annotation for C++ types that inherit from a base type annotated with ```SWIFT_SHARED_REFERENCE```. These derived types are also imported as a reference type in Swift that use the same ```retain``` and ```release``` functions as the base type. No additional annotation is required on the derived types.
+When a C++ type inherits from a `SWIFT_SHARED_REFERENCE` base type, the Swift compiler automatically infers `SWIFT_SHARED_REFERENCE` annotation for the derived type.
+The derived type also gets imported as a reference type, and uses the same `retain` and `release` functions as its base class.
 
 ### Inheritance and Virtual Member Functions
 
--- a/documentation/cxx-interop/index.md
+++ b/documentation/cxx-interop/index.md
@@ -1256,7 +1256,8 @@ object.doSomething()
 ```
 
 ### Inference for Derived Types
-Swift compiler automatically infers the ```SWIFT_SHARED_REFERENCE``` annotation for C++ types that inherit from a base type annotated with ```SWIFT_SHARED_REFERENCE```. These derived types are imported as Swift reference types or Swift class types that use the same ```retain``` and ```release``` functions as the base type. No additional annotation is required on the derived types.
+
+Swift compiler automatically infers the ```SWIFT_SHARED_REFERENCE``` annotation for C++ types that inherit from a base type annotated with ```SWIFT_SHARED_REFERENCE```. These derived types are also imported as a reference type in Swift that use the same ```retain``` and ```release``` functions as the base type. No additional annotation is required on the derived types.
 
 ### Inheritance and Virtual Member Functions
 
--- a/documentation/cxx-interop/index.md
+++ b/documentation/cxx-interop/index.md
@@ -1259,6 +1259,12 @@ object.doSomething()
 
 When a C++ type inherits from a `SWIFT_SHARED_REFERENCE` base type, the Swift compiler automatically infers `SWIFT_SHARED_REFERENCE` annotation for the derived type.
 The derived type also gets imported as a reference type, and uses the same `retain` and `release` functions as its base class.
+This inference works as long as all the annotated base types in the inheritance chain (including multiple or indirect inheritance) have the same `retain` and `release` functions.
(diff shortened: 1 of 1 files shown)


documentation/cxx-interop/index.md
```diff
@@ -1255,6 +1255,9 @@ object.doSomething()
 // `object` will be released here.
 ```
 
+### Inference for Derived Types
+Swift compiler automatically infers the ```SWIFT_SHARED_REFERENCE``` annotation for C++ types that inherit from a base type annotated with ```SWIFT_SHARED_REFERENCE```. These derived types are imported as Swift reference types or Swift class types that use the same ```retain``` and ```release``` functions as the base type. No additional annotation is required on the derived types.
```
Xazax-hun: What do you mean by Swift reference types or Swift class types? I think we should just pick one terminology here. 

Maybe we want to mention that this inference only happens if exactly one base type is annotated as a shared reference?

documentation/cxx-interop/index.md
```diff
@@ -1255,6 +1255,9 @@ object.doSomething()
 // `object` will be released here.
 ```
 
+### Inference for Derived Types
```
egorzhdan: Minor: missing newline after a heading

documentation/cxx-interop/index.md
fahadnayyar: > What do you mean by Swift reference types or Swift class types? I think we should just pick one terminology here.

Let's stick with **Swift reference types**, since that term is already used earlier in the doc (e.g., _“Now that SharedObject is imported as a reference type in Swift,…”_).

> Maybe we want to mention that this inference only happens if exactly one base type is annotated as a shared reference?

That’s not entirely accurate. In cases where a type inherits from multiple bases (e.g., due to diamond inheritance or transitive inheritance), Swift will still infer `SWIFT_SHARED_REFERENCE` as long as all annotated base types use the same `retain` and `release` functions.
A diagnostic is only emitted when there are conflicting `retain` and `release` function pairs among the base types.

fahadnayyar:
CC: @j-hui 
if you can also please take a look!

documentation/cxx-interop/index.md
```diff
@@ -1255,6 +1255,10 @@ object.doSomething()
...
+
+Swift compiler automatically infers the ```SWIFT_SHARED_REFERENCE``` annotation for C++ types that inherit from a base type annotated with ```SWIFT_SHARED_REFERENCE```. These derived types are also imported as a reference type in Swift that use the same ```retain``` and ```release``` functions as the base type. No additional annotation is required on the derived types.
+
 ### Inheritance and Virtual Member Functions
 
 Similar to value types, casting an instance of a derived reference type to a
```
fahadnayyar, line 1271: @egorzhdan, In these lines, are we referring to casting on the C++ side or the Swift side?

fahadnayyar:
CC: @ravikandhadai

documentation/cxx-interop/index.md
```diff
@@ -1255,6 +1255,10 @@ object.doSomething()
 // `object` will be released here.
 ```
 
+### Inference for Derived Types
```
j-hui: Does this even need a new heading? I don't think "Inference for Derived Types" is very clear or precise, but I'm not sure how much it actually adds to the single paragraph it summarizes. I think the couple of sentences you added are fine as a closing remark for the "Shared Reference Types" section.

documentation/cxx-interop/index.md
j-hui: I think Swift, i.e.,

> ... is not yet supported by Swift.

documentation/cxx-interop/index.md
```diff
@@ -1255,6 +1255,10 @@ object.doSomething()
...
+
+Swift compiler automatically infers the ```SWIFT_SHARED_REFERENCE``` annotation for C++ types that inherit from a base type annotated with ```SWIFT_SHARED_REFERENCE```. These derived types are also imported as a reference type in Swift that use the same ```retain``` and ```release``` functions as the base type. No additional annotation is required on the derived types.
```
j-hui: A couple of things:
- Use single backticks instead of triple backticks for inline code markup
- This doesn't say anything about multiple inheritance, which inference does not handle

I rephrased things a little to encompass those changes:

```suggestion
When a C++ class inherits from exactly one `SWIFT_SHARED_REFERENCE` base class, the Swift compiler automatically infers the derived class to be a `SWIFT_SHARED_REFERENCE` as well.
The derived type also gets imported as a reference type, and uses the same `retain` and `release` functions as its base class.
C++ classes that inherit from multiple `SWIFT_SHARED_REFERENCE` base classes must still be explicitly annotated with `retain` and `release` operations that manage the reference counts of all base classes.
```

documentation/cxx-interop/index.md
fahadnayyar: @ravikandhadai suggested adding a new heading for this feature. One reason why a new heading might be useful is that the inheritance feature doesn't apply to `SWIFT_IMMORTAL_REFERENCE` annotations.

I agree that this is a bad title. Maybe we can write **Inference of Shared Reference behaviour in Derived Types**

documentation/cxx-interop/index.md
fahadnayyar: > This doesn't say anything about multiple inheritance, which inference does not handle

> When a C++ class inherits from **exactly one** `SWIFT_SHARED_REFERENCE` base class, the Swift compiler automatically infers the derived class to be a `SWIFT_SHARED_REFERENCE` as well.

This is not entirely correct. It is fine to have multiple base `SWIFT_SHARED_REFERENCE` types. The only restriction is that all of them must have the same `retain`/`release` functions.

fahadnayyar:
This patch needs 2 important things to be included:
- behaviour in case of multiple inheritance.
- Clarification about immortal references.

documentation/cxx-interop/index.md
```diff
@@ -1255,6 +1255,11 @@ object.doSomething()
 // `object` will be released here.
 ```
 
+### Inference of Shared Reference behaviour in Derived Types
+
+When a C++ type inherits from a `SWIFT_SHARED_REFERENCE` base type, the Swift compiler automatically infers `SWIFT_SHARED_REFERENCE` annotation for the derived type.
+The derived type also gets imported as a reference type, and uses the same `retain` and `release` functions as its base class.
+
```
fahadnayyar, line 1268: @j-hui I changed some text based on your suggestions but added my flavor as well. How does it look now?

documentation/cxx-interop/index.md
fahadnayyar: CC: @j-hui

documentation/cxx-interop/index.md
```diff
@@ -1255,6 +1255,17 @@ object.doSomething()
 // `object` will be released here.
 ```
 
+### Inference of Shared Reference behaviour in Derived Types
```
fahadnayyar, line 1258: TODO: ### -> ####

fahadnayyar:
including this in [PR-1106](https://github.com/swiftlang/swift-org-website/pull/1106)

PR #1086, opened 2025-06-16
Added compatibility information for Wasm and Android to all packages

daveverwer:
### Motivation:

The Swift Package Index recently added Wasm and Android compatibility testing. This PR, and the corresponding changes to the [PackageListTool](https://github.com/SwiftPackageIndex/PackageListTool) automate the addition of this data into the Swift.org package listings.

### Modifications:

Adds Wasm and Android to package "cards" that are compatible with those platforms.

### Result:

The screenshot shows a variety of platform compatibilities, including one with all four that we can test for.

Screenshot 2025-06-16 at 18 03 49@2x

### DIFF
--- a/_data/packages/packages.yml
+++ b/_data/packages/packages.yml
@@ -30,7 +30,8 @@ categories:
         platform_compatibility:
           - Apple
           - Linux
-        platform_compatibility_tooltip: Apple (iOS, macOS, visionOS) and Linux
+          - Android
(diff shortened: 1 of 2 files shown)


daveverwer:
As discussed at the latest SWWG meeting, this has now been updated to only include Wasm and not Android due to Wasm being a fully supported platform now. 

The code is still in the PackageListTool for Android and can be easily re-enabled at any time.

The look and feel is the same, just without the Android compatibility:

Screenshot 2025-10-21 at 18 08 21@2x

PR #1065, opened 2025-06-06
Copy edit front page

finagolfin:
Light edit of new front page copy, some of it matter of taste.

PR #1093, opened 2025-06-22
Fix wrong marketing strings

soapdog:
Fix misleading marketing message. 

### Motivation:

All the motivation is explained in https://github.com/swiftlang/swift-org-website/issues/1044 The website contains false information about Swift being the only language capable of doing kernels, apps, and cloud. 

### Modifications:

Fixed the misleading strings.

### Result:

The website will no longer contain claims about Swift that are proven to be false.

### DIFF
--- a/_data/new-data/landing/callouts.yml
+++ b/_data/new-data/landing/callouts.yml
@@ -71,7 +71,7 @@
 
 - title: Adaptable
   subtitle: From microcontrollers to servers.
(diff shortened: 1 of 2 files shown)


davelester:
Thanks for the contribution, @soapdog! The homepage text was revised in #1369 and #1065, which replaced the text mentioned in this PR. Closing this out, but thank you for taking the time!

PR #1109, opened 2025-07-09
Fix "Copy install command" button on old releases page

chromy:
There are two definitions of `copyToClipboard` in the codebase, one with one arg
and one with two args. The onClick handler ends up calling the wrong one for a
number of links. I assume the issue is something like:
- Both definitions end up being defined on the same page
- The onClick ends up getting the one with two args when it expects the one with one arg.

Attempt to fix this by instead using the two arg version and removing the one arg version.
However I do not have a Ruby environment setup to test with so I haven't been able to try it.

This resolves https://github.com/swiftlang/swift-org-website/issues/1108.


shahmishal (approved):

heckj (approved):
Verified with a locally run version as well

PR #1110, opened 2025-07-10
Add directory for info architecture proposals and first two proposals

dempseyatgithub:
As discussed in the last website workgroup meeting, Swift info architecture project proposals and documents will be included in the swift-org-website repository.

The proposals are *not* rendered as webpages on swift.org.

Jekyll automatically ignores directories and files beginning with `#`. Since files beginning with an underscore are typically part of the configuration / operation of Jekyll itself, the directory name is `#info-architecture` to emphasize this is not part of the Jekyll infrastructure or published on the website.

This PR includes the first two proposals that have been announced and reviewed by the community. Future proposals will be added when they are ready for community review and feedback.

### DIFF
--- a/_info-architecture/0001-swift-information-architecture-project.md
+++ b/_info-architecture/0001-swift-information-architecture-project.md
@@ -1,0 +1,220 @@
+# Swift Information Architecture Project
+
+* Proposal: [SIAP-0001](0001-swift-information-architecture-project.md)
+* Author: [James Dempsey](https://github.com/dempseyatgithub)
+
+## Overview
+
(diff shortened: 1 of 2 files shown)


heckj (changes requested):
Content is great, but would you mind terribly if we switched the prefix from `#` to `_`, as the `#` symbol is causing no end of pain for me when I'm trying to interact with it on the command line. 

Plus, that keeps in line with the other "ignored" directories that need to be explicitly included per config (https://jekyllrb.com/docs/structure/)

PR #1111, opened 2025-07-14
mentorship: Move year from timeline subheader to parent header

AnthonyLatsis:
### Motivation:

We link to the timeline in email messages to participants. Having the year in the timeline subheader makes it very easy to forget to update the URL fragment each year.


AnthonyLatsis:
We will link to https://www.swift.org/mentorship/#current-program instead.

PR #1129, opened 2025-08-12
Re-order the server packages

FranzBusch:
The order of the server packages was a bit arbitrary. This PR re-orders them by sorting them from highest to lowest level, for their respective importance and then how production ready they are.

### DIFF
--- a/_data/new-data/get-started/cloud-services/tertiary-content.yml
+++ b/_data/new-data/get-started/cloud-services/tertiary-content.yml
@@ -7,15 +7,15 @@ tertiary_cards:
     text: Generate Swift client and server code from an OpenAPI document.
     link_text: View on GitHub
     link: https://github.com/apple/swift-openapi-generator
-  - name: SwiftNIO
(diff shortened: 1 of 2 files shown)


_data/new-data/get-started/cloud-services/tertiary-content.yml
```diff
@@ -7,15 +7,15 @@ tertiary_cards:
     text: Generate Swift client and server code from an OpenAPI document.
     link_text: View on GitHub
     link: https://github.com/apple/swift-openapi-generator
-  - name: SwiftNIO
-    text: Cross-platform asynchronous event-driven network application framework.
+  - name: Async HTTP Client
```
0xTim, line 10: You said you reordered this from most to important at the top, but NIO is above AHC right?

FranzBusch: I re-ordered them from highest in the stack to lowest in the stack first, then importance and then matureness. That's why NIO is last here because most users should not go to NIO directly but rather one of the top level packages first.

PR #1133, opened 2025-08-16
Use a curlbash script for installing swiftly

marcprux:
### Motivation:

The current recommendation for installing Swiftly on [macOS](https://www.swift.org/install/macos/) and [Linux](https://www.swift.org/install/linux/) involves a gnarly multi-line shell-specific blob of code that the user needs to paste into their shell. 

Curlbash scripts (i.e., `curl https://… | sh`) are a common and portable way to install software for UNIX-like systems. They are used by popular packages like [rbenv](https://rbenv.org), [Homebrew](https://brew.sh), [Rust](https://www.rust-lang.org/tools/install), [NVM](https://github.com/nvm-sh/nvm#installing-and-updating), [Tailscale](https://tailscale.com/kb/1031/install-linux), and [Deno](https://docs.deno.com/runtime/getting_started/installation/).

### Modifications:

Add a `swiftly-install` shell script for installing swiftly and recommend using it from the "Install" pages.

### Result:

The installation instructions will be briefer and nicer and not require that the user know what shell they are running.

Before:

Screenshot 2025-08-16 at 18 03 08

After:

Screenshot 2025-08-16 at 18 03 23

### DIFF
--- a/swiftly-install
+++ b/swiftly-install
@@ -1,0 +1,38 @@
+#!/bin/bash -e
+# Install script for Swiftly
+# Usage: curl -fsSL https://swift.org/swiftly-install | sh
+OS_NAME=$(uname -s)
+OS_ARCH=$(uname -m)
+TMPDIR=$(mktemp -d)
+
+cd ${TMPDIR}
+
+case "$OS_NAME" in
+    "Linux")
+        curl -fsSLO https://download.swift.org/swiftly/linux/swiftly-${OS_ARCH}.tar.gz
+        tar zxf swiftly-${OS_ARCH}.tar.gz
+        ./swiftly init
+        SWIFTLY_HOME_DIR=${SWIFTLY_HOME_DIR:-$HOME/.swiftly}
+        ;;
+    "Darwin")
+        curl -fsSLO https://download.swift.org/swiftly/darwin/swiftly.pkg
+        installer -pkg swiftly.pkg -target CurrentUserHomeDirectory
+        ~/.swiftly/bin/swiftly init
+        SWIFTLY_HOME_DIR=${SWIFTLY_HOME_DIR:-$HOME/.local/share/swiftly}
+        ;;
+    *)
+        echo "Unknown platform: $OS_NAME"
+        echo "This script supports Linux and Darwin/macOS only"
+        exit 1
+        ;;
+esac
+
+echo "Swiftly installed! Run the following command to set up your environment:"
+PARENT_SHELL=$(ps -o comm= -p "$PPID")
+if [ "$PARENT_SHELL" = "fish" ]; then
+    echo "source ${SWIFTLY_HOME_DIR}/env.fish"
+else
+    # Not fish: assume sh-compatible (bash/zsh)
+    echo "source ${SWIFTLY_HOME_DIR}/env.sh"
+fi
+
--- a/_data/new-data/install/linux/releases.yml
+++ b/_data/new-data/install/linux/releases.yml
@@ -3,20 +3,7 @@ latest-release:
     pre-code-text: |
       The Swiftly installer manages Swift and its dependencies. It supports switching between different versions and downloading updates.
     headline: Swiftly
-    tabs:
-      - label: Bash
-        code: |-
-          curl -O https://download.swift.org/swiftly/linux/swiftly-$(uname -m).tar.gz && \
-          tar zxf swiftly-$(uname -m).tar.gz && \
-          ./swiftly init --quiet-shell-followup && \
-          . "${SWIFTLY_HOME_DIR:-$HOME/.local/share/swiftly}/env.sh" && \
-          hash -r
-      - label: Fish
-        code: |-
-          curl -O https://download.swift.org/swiftly/linux/swiftly-(uname -m).tar.gz && \
-          tar zxf swiftly-(uname -m).tar.gz && \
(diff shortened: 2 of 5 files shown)


swiftly-install
```diff
@@ -0,0 +1,52 @@
+#!/bin/bash
```
jakepetroules: Use `#!/bin/sh` for maximum portability? `shellcheck` can help validate.

marcprux: soundness.sh seems to prefer bash

swiftly-install
```diff
@@ -0,0 +1,52 @@
...
+## Copyright (c) 2025 Apple Inc. and the Swift.org project authors
+## Licensed under Apache License v2.0
+##
+## See LICENSE.txt for license information
+## See CONTRIBUTORS.txt for the list of Swift.org project authors
+##
+## SPDX-License-Identifier: Apache-2.0
+##
+##===----------------------------------------------------------------------===##
+
```
jakepetroules:
```suggestion
set -eu
```

swiftly-install
```diff
@@ -0,0 +1,52 @@
...
+# Install script for Swiftly
+# Usage: curl -fsSL https://swift.org/swiftly-install | sh
+OS_NAME=$(uname -s)
+OS_ARCH=$(uname -m)
+TMPDIR=$(mktemp -d)
+
+cd ${TMPDIR}
```
jakepetroules:
```suggestion
cd "${TMPDIR}"
```

swiftly-install
```diff
@@ -0,0 +1,52 @@
...
+
+case "$OS_NAME" in
+    "Linux")
+        curl -fsSLO https://download.swift.org/swiftly/linux/swiftly-${OS_ARCH}.tar.gz
```
jakepetroules:
```suggestion
        curl -fsSLO "https://download.swift.org/swiftly/linux/swiftly-${OS_ARCH}.tar.gz"
```

swiftly-install
```diff
@@ -0,0 +1,52 @@
...
+        tar zxf swiftly-${OS_ARCH}.tar.gz
```
jakepetroules:
```suggestion
        tar zxf "swiftly-${OS_ARCH}.tar.gz"
```

swiftly-install
```diff
@@ -0,0 +1,52 @@
...
+# Install script for Swiftly
+# Usage: curl -fsSL https://swift.org/swiftly-install | sh
+OS_NAME=$(uname -s)
+OS_ARCH=$(uname -m)
+TMPDIR=$(mktemp -d)
+
+cd ${TMPDIR}
+
+case "$OS_NAME" in
+    "Linux")
```
jakepetroules, line 31: On Linux, almost no command line tool is guaranteed to be installed. It's worth checking if all of the programs the script needs are installed.

- `ps` (from a package usually called procps) is not always present.
- `tar` is not always present.
- Neither is curl (yes the swift.org instructions propose fetching the script with curl in the first place, but someone might've still manually used wget or something other way to get the script onto their machine)

mktemp and uname come from coreutils, but those are a bit too fundamental to bother checking for IMO. cd is a shell builtin.

Or, you could detect apt-get and dnf and try to install the relevant packages to ensure they'll be there (without `-y`, so the user is prompted if configured).

something like this:

```sh
if command -v apt-get >/dev/null 2>&1 ; then # debian, ubuntu
    apt-get update
    apt-get install curl procps tar
elif command -v dnf >/dev/null 2>&1 ; then # rhel
    dnf update
    dnf install curl procps tar
elif command -v yum >/dev/null 2>&1 ; then # amazonlinux2
    yum update
    yum install curl procps tar
elif ! command -v curl >/dev/null 2>&1 || ! command -v ps >/dev/null 2>&1 || ! command -v tar >/dev/null 2>&1 ; then
    echo "Missing one of more of the following programs: curl, ps, tar" >&2
    exit 1
fi
```

marcprux: Good call. I think that automatically running their package manager to install the missing utilities is a bit heavy-handed, so I just went with the fail-fast exit when one of the necessary commands is not present.

marcprux:
Has there been any further consideration of this? I think it would make a very nice improvement to the installation experience.

I just noticed that @theolampert uses a similar technique in https://github.com/theolampert/swift-android-build-action/blob/main/action.yml which has been adopted by a number of GitHub workflows. It'd be nice to centralize on one authoritative way of doing it.

marcprux:
> I think @airspeedswift had some security concerns with this approach (while others do it, is it really a good practice to encourage people to pipe unread shell scripts to their interpreter)?

The current mechanism also involves running an unread script obtained by curl, it just does it in a much more cumbersome way than the proposed one-liner. Happy to discuss further, but no matter how it is conveyed, we are ultimately asking the user to execute some arbitrary code downloaded from swift.org — either they trust that, or they don't.

This was how swiftly used to recommend installation, back when it was located at https://swiftlang.github.io/swiftly/.

> That said, I do agree that this is vastly more developer-friendly. I wonder if there's a way we could link to the script that will be run, so folk can review it before executing it? Don't know if that's enough to satisfy the security concern?

They could just run the curl command _without_ piping it through `sh` to see the contents (which is what I always do before executing any curlbash elsewhere on the web). We could additionally provide a link to the script if we think that would help.

PR #1140, opened 2025-09-01
[windows] update the install instructions to use Python 3.10

charles-zablit:
Update the install instructions to use Python 3.10.

### Motivation:

https://github.com/swiftlang/swift/pull/83929 updated the Windows toolchain to use Python 3.10.1. Python 3.10.x is now required to use lldb on Windows.

### Modifications:

Update the install instructions to use Python 3.10 instead of 3.9.

### Result:

Users will be instructed to install Python 3.10 instead of 3.9.

rdar://159596578

### DIFF
--- a/_includes/install/_windows_dependencies.md
+++ b/_includes/install/_windows_dependencies.md
@@ -5,7 +5,7 @@ Swift has the following general dependencies:
 - Git (used by Swift Package Manager)
 - Python[^1] (used by the debugger - LLDB)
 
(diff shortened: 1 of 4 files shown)


heckj (approved):

shahmishal:
This will only work after 6.2 has been released, 6.1.x will still require 3.9.

charles-zablit:
Should we revert the change and wait for `6.2.1` to be released?

PR #1186, opened 2025-10-08
Redirect swift server concurrency guide content to the Swift 6 migration guide

heckj:
### Motivation:

The content in this article is identical to the content in the Swift 6 Migration Guide, so to consolidate to a single location to update about migrations and concurrency, I'm redirecting to the data race safety article in the migration guide.

### Modifications:

Keep front-matter and redirect page content to the Migration Guide book/content

### Result:

Any access to the page will automatically redirect to the migration guide.

### DIFF
--- a/documentation/concurrency/index.md
+++ b/documentation/concurrency/index.md
@@ -2,9 +2,9 @@
 layout: page
 date: 2024-03-03 12:00:00
 title: Enabling Complete Concurrency Checking
-redirect_to: https://www.swift.org/migration/documentation/swift-6-concurrency-migration-guide/enabledataracesafety
+redirect_to: /migration/documentation/swift-6-concurrency-migration-guide/enabledataracesafety/
 ---
-<meta http-equiv="refresh" content="0; url=https://www.swift.org/migration/documentation/swift-6-concurrency-migration-guide/">
-<link rel="canonical" href="https://www.swift.org/migration/documentation/swift-6-concurrency-migration-guide/" />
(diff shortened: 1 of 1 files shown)


documentation/concurrency/index.md
```diff
@@ -2,63 +2,9 @@
 layout: page
 date: 2024-03-03 12:00:00
 title: Enabling Complete Concurrency Checking
+redirect_to: https://www.swift.org/migration/documentation/swift-6-concurrency-migration-guide/enabledataracesafety
 ---
+<meta http-equiv="refresh" content="0; url=https://www.swift.org/migration/documentation/swift-6-concurrency-migration-guide/">
+<link rel="canonical" href="https://www.swift.org/migration/documentation/swift-6-concurrency-migration-guide/" />
```
alexandersandberg: Do we want to redirect to `https://www.swift.org/migration/documentation/swift-6-concurrency-migration-guide/enabledataracesafety` or `https://www.swift.org/migration/documentation/swift-6-concurrency-migration-guide/` here? I believe the former?

Also, it seems like the latter would be `https://www.swift.org/migration/documentation/migrationguide`. I'm getting a forbidden page otherwise.

alexandersandberg (commented):
Agree on the redirect, but see my comment below.

documentation/concurrency/index.md
heckj: Just updated and rebased!

alexandersandberg (approved):

PR #1232, opened 2025-11-12
Add support for Apple `container` for building and running website locally

timsneath:
Reduces friction for hosting the website locally on macOS.

### Motivation:

[Container](https://github.com/apple/container) is a lightweight, free tool, written in Swift, that enables secure local hosting of Linux containers on macOS. For website contributors using macOS, this is a low-friction way to build and host the website without any other system dependencies. 

### Modifications:

- Added a Makefile that provides actions for building and serving the image, as well as cleaning up artifacts
- Added instructions to the README to install and use `container` for website hosting.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -1,23 +1,30 @@
-# Makefile for managing the Swift website with the container app
+# Makefile for managing the Swift website with container
+
+# Define the container runtime. Default to `container`.
+# Can be overridden from the command line, e.g., `make CONTAINER=docker website`
+CONTAINER ?= container
 
 .PHONY: help build run-build website stop clean
 
 help:
 	@echo "Usage:"
-	@echo "  make build    Build the swift-website-builder container image"
+	@echo "  make build      Build the swift-website-builder container image"
 	@echo "  make run-build  Build the Jekyll website"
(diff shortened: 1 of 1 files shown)


heckj:
I'm having a devil of a time getting Ruby 3.3.x (9 as if this morning) to compile locally, so 👍 from me!

Makefile
```diff
@@ -0,0 +1,44 @@
...
+	@echo "  make clean    Stop the container and remove the build output"
+
+# Build the primary container image
+build:
+	container build --tag swift-website-builder --file Dockerfile .
+
+# Run a one-off Jekyll build
+run-build:
+	@mkdir -p ./.output
+	container run --rm \
```
jglogan: Since we're not using any special CLI constructs here, what if we defined a make variable $(CONTAINER) that defaults to `container`, but `make CONTAINER=docker target` would work fine as well. Or `podman`, `nerdctl`, etc.

jglogan (approved):
See note but otherwise lgtm

Makefile
timsneath: Great! Updated this to be tool-agnostic.

PR #1252, opened 2025-12-01
Remove old prerelease caveats from cloud packages page

timsneath:
Remove "pre-1.0" warnings from OTel and OPA, since both are now published. 

- https://forums.swift.org/t/swift-otel-1-0-0-released/82439
- https://blog.openpolicyagent.org/introducing-swift-opa-native-policy-evaluation-for-swift-d5136c8a662e


_data/new-data/get-started/cloud-services/tertiary-content-cloud-native.yml
```diff
@@ -5,14 +5,14 @@ tertiary_cards:
...
+    text: OpenTelemetry client built for Swift observability libraries.
     link_text: View on GitHub
     link: https://github.com/swift-otel/swift-otel
   - name: Swift Prometheus
     text: Client for the Prometheus monitoring system, supporting counters, gauges and histograms.
     link_text: View on GitHub
     link: https://github.com/swift-server/swift-prometheus
   - name: Swift OPA
-    text:  Evaluate Open Policy Agent IR Plans compiled from Rego declarative policy. (pre-1.0 release)
+    text:  Evaluate Open Policy Agent IR Plans compiled from Rego declarative policy.
```
czechboy0: Swift OPA still is pre-1.0: https://github.com/open-policy-agent/swift-opa

timsneath: Oh, interesting. I know you're authoritative :) but I couldn't find any reference to it being still a pre-release. 

- Neither https://www.openpolicyagent.org/ecosystem/entry/swift-opa nor https://www.openpolicyagent.org/ecosystem mention this
- https://blog.openpolicyagent.org/introducing-swift-opa-native-policy-evaluation-for-swift-d5136c8a662e says it's "now been released". 
- https://github.com/open-policy-agent/swift-opa?tab=readme-ov-file#usage doesn't mention anything about it being not ready for production usage. 

In any event, I wonder if this really is the place to track its status -- any of the above URLs would probably a better place to describe its status, since they're the places that will get updated with a change?

czechboy0: The project being pre-1.0 just means it doesn't have a stable API yet, so packages can't depend on it expecting a stable API for a few years.

Swift OPA doesn't appear to have any releases yet (https://github.com/open-policy-agent/swift-opa/releases), so the only way to depend on it would be `branch: "main"` which would prevent your package from being even tagged with a version, I believe (SwiftPM would reject it).

I think of production readiness as an orthogonal metric, some packages are solid even before 1.0, others need time to bake post-1.0. But our original note about the lack of 1.0 was to call out the lack of a stable API yet.

Whether we want to keep it I'm agnostic about, fine either way.

shahmishal:
https://github.com/swift-otel/swift-otel is now at 1.5.0 version, going to merge this PR.

heckj (approved):

heckj:
(added a commit to fix the soundness issue that the YAMLLint bit was complaining about - a trailing whitespace trimmed.

PR #1253, opened 2025-12-04
Fix error message syntax in static Linux Swift SDK guide

MaxDesiatov:
Per [user's report](https://forums.swift.org/t/cross-compiling-for-linux/83523/9) `error(Unknown platform)` is invalid syntax. The error message string should be quoted.

### DIFF
--- a/documentation/articles/static-linux-getting-started.md
+++ b/documentation/articles/static-linux-getting-started.md
@@ -254,7 +254,7 @@ Hello, world!
 
 ### What about package dependencies?
 
-Swift packages that make use of Foundation or Swift NIO should just
+Swift packages that make use of Foundation or SwiftNIO should just
 work.  If you try to use a package that uses the C library, however,
 you may have a little work to do.  Such packages often contain files
 with code like the following:


al45tair (approved):

documentation/articles/static-linux-getting-started.md
```diff
@@ -254,7 +254,7 @@ Hello, world!
 
 ### What about package dependencies?
 
-Swift packages that make use of Foundation or Swift NIO should just
+Swift packages that make use of Foundation or SwiftNIO should just
```
finagolfin, line 257: I think separating the two is better, what do you think, @Lukasa?

MaxDesiatov: Separating what exactly?

MaxDesiatov: SwiftNIO [is the established name of the project](https://github.com/apple/swift-nio?tab=readme-ov-file#swiftnio). There's no question of "better" or "worse" for an established name, it's just following what the project authors intended for correct writing, in the same way we don't write "mac OS", "Swift PM", or "Git Hub".

finagolfin: OK, didn't see that, it's a bad name they chose then.

PR #1254, opened 2025-12-04
Display Swift 6.3 snapshots on swift.org/install page

MaxDesiatov:
We started publishing Swift 6.3 development snapshots for most platforms, so those should take precedence over 6.2 snapshots.


finagolfin:
Looks good, I only see WASM SDK snapshots added here though, anything holding up the other 6.3 SDKs?

shahmishal:
We are going to close this in favor of a new one I am creating.

PR #1272, opened 2025-12-16
Add smooth scroll for anchor click

federicobucchi:
### Motivation:

When the user clicks on an anchor the scroll is a straight jump.
It would be a nicer experience having a gentle / smooth scroll.

### Modifications:

Add the smooth scroll for `html` taking in consideration the accessibility reduce motion.

### Result:

Smooth scroll is applied (if not disabled by the accessibility feature reduce motion)


heckj (approved):
subtle!

PR #1283, opened 2026-01-09
Change Contributor role nomination to send message on forums instead of email

ahoppen:
We have seen reports in the past (eg. https://forums.swift.org/t/a-new-member-contributor-role-that-can-trigger-ci/78074/9) that you can’t send emails to the code-owners group. Sending messages on the Swift forums should work, though. Re-phrase the nomination process to send a message on the forums instead of sending an email. This also means that the contributor has a Swift Forum account, which may be useful to contact them in the future.


parispittman:
lgtm

PR #1313, opened 2026-02-03
Add prerequisites section to Windows install page

### LINKED ISSUE #763: Windows install instructions should make it obvious that you need to install Visual Studio
[The Windows install page](https://www.swift.org/install/windows/) does not make it sufficiently obvious that installing Visual Studio is a prerequisite for a working Swift install. If you're using `winget`, of course, this happens automatically and there is no problem, but if you are using the GUI installer, which is what most people will reach for by default, that is not the case.

Yes, for the release version, there is an instructions link just under the installer link, but there isn't anything for the nightly snapshots and there is no indication on the page itself that just installing using the GUI installer will not be sufficient (or that using `winget` might be advantageous, in spite of requiring the use of the command prompt).

The symptoms you get when you try to use Swift without Visual Studio installed are perplexing for a new user of Swift; you see errors like

```
26 | // __UINTxx_TYPE__ are defined only since Clang 3.5.
27 | #if !defined(__APPLE__) && !defined(__linux__) && !defined(__OpenBSD__) && !defined(__wasi__)
28 | #include <stdint.h>
   |          `- error: 'stdint.h' file not found
29 | typedef int64_t __swift_int64_t;
30 | typedef uint64_t __swift_uint64_t;
```

when trying to build a simple program (even the default generated by `swift package init --type executable`, actually), and the cause is not obvious.

guilhermemouraovc:
Add prerequisites section to Windows install page

Make Visual Studio requirement more prominent by adding a dedicated prerequisites section at the top of the Windows installation page. This helps prevent build errors from users installing Swift without required dependencies.

### Motivation:

The Windows install page did not make it sufficiently obvious that Visual Studio is a required prerequisite for Swift. Users who missed this requirement encountered cryptic build errors like 'stdint.h' file not found when attempting to compile Swift code. Issue #666 documented this problem where new users could easily skip the Visual Studio installation, especially when using the manual installer route.

### Modifications:

- Added a prominent "Prerequisites" section at the top of /install/windows/index.md before the installation steps
- Included the existing _windows_dependencies.md content that lists all required dependencies (Git, Python, Windows SDK, Visual Studio)
- Added an important note clarifying that WinGet automatically installs Visual Studio, but manual installer users must install prerequisites separately
- Updated the section 1 heading from "Install Swift via WinGet" to "Install Swift and Dependencies via WinGet" to make it clear both are included

### Result:

Users visiting the Windows installation page will now immediately see the prerequisites section with Visual Studio clearly listed as a required dependency. This prevents confusion and reduces support burden from users who attempt to use Swift without the necessary platform dependencies installed.

Fixes #763

### DIFF
--- a/install/windows/index.md
+++ b/install/windows/index.md
@@ -13,7 +13,14 @@ title: Install Swift - Windows
 {% assign platform = site.data.builds.swift_releases.last.platforms | where: 'name', 'Windows 10' | first %}
 
 <div class="content">
-  <h3 id="winget" class="header-with-anchor">1. Install Swift via WinGet</h3>
+  <div class="release-box section">
+    <div class="content">
+      <h2 id="prerequisites" class="header-with-anchor">Prerequisites</h2>
+      {% include install/_windows_dependencies.md %}
+      <p class="body-copy"><strong>Important:</strong> The WinGet command in step 1 below will automatically install Visual Studio for you. If you prefer to use the manual installer, you must install these prerequisites separately before installing Swift.</p>
+    </div>
+  </div>
+  <h3 id="winget" class="header-with-anchor">1. Install Swift and Dependencies via WinGet</h3>
   <div class="release-box section">
     <div class="content">
       {% include new-includes/components/code-box.html content = site.data.new-data.install.windows.releases.latest-release.winget %}


shahmishal:
cc: @al45tair

Issue #1268, opened 2025-12-15
Error in "Value and Reference Types" article: class Document missing initializer


kyleqihua:
In the Swift.org article "Value and Reference Types in Swift" (https://www.swift.org/documentation/articles/value-and-reference-types.html), the reference-type example uses a class `Document` defined as:

```swift
class Document {
    var text: String
}
var myDoc = Document(text: "Great new article")
var friendDoc = myDoc
friendDoc.text = "Blah blah blah"
print(friendDoc.text)
print(myDoc.text)
```

In Swift, classes do **not** receive a synthesized memberwise initializer. Because the stored property `text` has no default value, the compiler does not automatically generate an initializer, so the call `Document(text: "Great new article")` fails to compile with an error such as "Class 'Document' has no initializers". To make this example compile, either:

- Provide a default value for `text`, e.g. `var text: String = ""` and then use `Document()`, or
- Define an explicit initializer: `init(text: String) { self.text = text }`.

As written, the code misleads readers into thinking that classes get memberwise initializers like structs. It would be helpful to update the article to either add an initializer or default value in the class example, or clarify this difference.

mpapple-swift:
Created PR #1315 to address this issue.

PR #1315, opened 2026-02-11
Fix Swift examples for value and reference types swiftlang issue #1268

mpapple-swift:
Fix examples to clarify value and reference types in Swift to address issue #1268.

 Change to value-and-reference-types.md to address a problem raised in swiftlang issue #1268 

### Motivation:

As mentioned in the associated swiftlang issue #1268, the 'reference' example is incorrect in that classes do not have an implicit initializer as is the case for structs.

### Modifications:

As suggest in the issue, I modified the class Document definition to add an initializer and set the value of the String variable 'text' to an empty string.
I have added some additional boilerplate in the form of classes that further the example given.  I am of the opinion that someone learning Swift from this article should be able to copy and paste the example code into Xcode or an IDE and have the code compile and run.  However, the existing example code, while clearly demonstrating the difference between value and reference types, will fail to compile as the print functions will produce a "Expressions are not allowed at the top level" error.  I am not wedded to these additions and will revert them to just modify the problem raised in issue #1268, but I hope you will consider the ergonomic aspect of providing code that compiles and runs as opposed to example code that merely demonstrates a point.

### Result:

The modifications to value-and-reference-types.md will address the fact that the reference example with a class type presents an incorrect usage of class semantics.

### DIFF
--- a/documentation/articles/value-and-reference-types.md
+++ b/documentation/articles/value-and-reference-types.md
@@ -42,14 +42,21 @@ Since it is an independent instance, changing the `text` of `friendDoc` does not
 struct Document {
   var text: String
 }
+```
+Example
+```swift
+class UseStructDocument {
+  var myDoc = Document(text: "Great new article")
 
-var myDoc = Document(text: "Great new article")
-var friendDoc = myDoc
+  init() {
+    var friendDoc = myDoc
 
-friendDoc.text = "Blah blah blah"
+    friendDoc.text = "Blah blah blah"
 
-print(friendDoc.text) // prints "Blah blah blah"
-print(myDoc.text) // prints "Great new article"
+    print(friendDoc.text) // prints "Blah blah blah"
+    print(myDoc.text) // prints "Great new article"
+  }
+}
 ```
 
 When you send your friend a copy of a document, you are in complete control of when *your* copy changes. You never have to worry about your friend making some unexpected change to your copy of the document.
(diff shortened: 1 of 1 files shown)


heckj (approved):

documentation/articles/value-and-reference-types.md
```diff
@@ -42,14 +42,21 @@ Since it is an independent instance, changing the `text` of `friendDoc` does not
 struct Document {
   var text: String
 }
+```
```
dempseyatgithub, line 45: As mentioned in my comment on the PR itself, please revert this section back to the original without the `UseStructDocument` type.

documentation/articles/value-and-reference-types.md
```diff
@@ -73,16 +80,26 @@ Since it is a reference to the same instance, changing the `text` of `friendDoc`
 
 ```swift
 class Document {
-  var text: String
-}
+  var text: String = ""
```
dempseyatgithub, line 83: Since the added `init` method takes and sets the text property, there is no need to initialize it here. Please revert this.

documentation/articles/value-and-reference-types.md
```diff
@@ -73,16 +80,26 @@ Since it is a reference to the same instance, changing the `text` of `friendDoc`
...
-var myDoc = Document(text: "Great new article")
-var friendDoc = myDoc
+  init(text: String) {
+    self.text = text
+  }
+}
+```
+Example
+```swift
+class UseClassDocument {
```
dempseyatgithub, line 92: As mentioned in my comment on the PR itself, please revert this section back to the original without the `UseClassDocument` type.

documentation/articles/value-and-reference-types.md
```diff
@@ -73,16 +80,26 @@ Since it is a reference to the same instance, changing the `text` of `friendDoc`
 
 ```swift
 class Document {
-  var text: String
-}
+  var text: String = ""
 
-var myDoc = Document(text: "Great new article")
-var friendDoc = myDoc
+  init(text: String) {
```
dempseyatgithub, line 85: The missing `init` method - yes, please keep.

dempseyatgithub (changes requested):
My rationale for requested changes is in the main PR discussion.

dempseyatgithub:
This pull request has been superseded by #1467. Thank you @mpapple-swift for this pr and starting the discussion about how to best update this article.

PR #1322, opened 2026-02-19
Update the zero-to-swift-nvim documentation

da-viper:
Some of the steps are deprecated and may throw an error. beginners may be confused that the must have missed a step in the article.

### Motivation:
I want to update the documentation to to setup the debugger. 

### Modifications:
Updating the current configuration to support new versions of neovim.

### Result:
There are are no errors when the user is setting up swift in neovim.

### DIFF
--- a/documentation/articles/zero-to-swift-nvim.md
+++ b/documentation/articles/zero-to-swift-nvim.md
@@ -176,8 +176,10 @@ To check that it's working:
 ## Language Server Support
 
 Language servers respond to editor requests providing language-specific support.
-Neovim has support for Language Server Protocol (LSP) built-in, so you don't
-need an external package for LSP. Neovim looks for lsp configurations in `<runtimepath>/lsp`. For more information run `:help lsp` and `:help rtp` in neovim. 
+Neovim has support for Language Server Protocol (LSP) built-in.
+Your configuration for each language server is installed to
+`<runtimepath>/lsp` and enabled by calling `vim.lsp.enable`.
+For more information run `:help lsp` and `:help rtp` in neovim.
 
 Go ahead and create a new file under `nvim/lsp/sourcekit.lua` which is a runtimepath. In it, we'll start by adding the following snippet. 
 


documentation/articles/zero-to-swift-nvim.md
etcwilde: It might be worth re-working this section entirely or splitting it between pre-`0.11.0` and post, since they made some pretty big changes to the LSP system in that version.

etcwilde: I'm not sure whether it's worth downloading this package anymore for a single language server.
From my own [dotfiles](https://github.com/etcwilde/dotfiles/blob/main/nvim/lsp/sourcekit.lua), this file should be completely sufficient for setting it up (in `<rtp>/lsp/sourcekit.lua`):

```lua
return {
  cmd = { 'sourcekit-lsp' },
  filetypes = { 'swift' },
  root_markers = {
    '.git',
    'compile_commands.json',
    '.sourcekit-lsp',
    'Package.swift',
  },
  get_language_id = function(_, ftype)
    return ftype
  end,
  capabilities = {
    workspace = {
      didChangeWatchedFiles = {
        dynamicRegistration = true,
      },
    },
    textDocument = {
      diagnostic = {
        dynamicRegistration = true,
        relatedDocumentSupport = true,
      },
    },
  },
}
```

Then I think it would be fine to reference `neovim/nvim-lspconfig` as a place to look up servers or to use for providing initial configs for many servers.

In either case, we should be able to drop the `didChangeWatchedFiles` if we're updating this because that is now included in `nvim-lspconfig`: https://github.com/neovim/nvim-lspconfig/blob/b7dcde3e6195b43476ae799720b177aedc16d44c/lsp/sourcekit.lua#L27-L39

etcwilde (commented):
The lazy install changes look good to me. The lsp implementation and configuration was overhauled pretty dramatically in nvim 0.11.0. If we're updating this document for 0.11+, I think there are more refinements we could make. I tried to choose a minimal set of plugins when originally writing this document.

Since we're dropping the `require('lspconfig')`, which is reasonable because it was deprecated as a package, I have mixed thoughts on whether it's worth pulling in every lsp server config, or should just show our own.

These changes seem fine, though I make take another pass at the lsp configuration section in the future.

documentation/articles/zero-to-swift-nvim.md
da-viper: I think we should remove any documentation for pre 0.11.0 because. 0.11.0 was released approximately 1 year ago. and most Neovim users will likely be on the latest stable version. Applies to new Neovim users as they are more likely to install the latest version.

da-viper: Yeah I works without the `neovim/nvim-lspconfig` plugin. will use that instead.

documentation/articles/zero-to-swift-nvim.md
```diff
@@ -178,25 +177,60 @@ To check that it's working:
 
 Language servers respond to editor requests providing language-specific support.
 Neovim has support for Language Server Protocol (LSP) built-in, so you don't
-need an external package for LSP, but adding a configuration for each LSP server
```
etcwilde, line 181: I think we can drop the reference to an external plugin here and just say how to enable it and where the configurations go.
Maybe something like this?

> Neovim has support for Language Server Protocol (LSP) built-in.
Your configuration for each language server is installed to `<runtimepath>/lsp` and enabled
by calling `vim.lsp.enable`.
For more information run `:help lsp` and `:help rtp` in neovim.

shahmishal (approved):

PR #1349, opened 2026-03-10
Fix the download.swift.org OpenAPI doc

czechboy0:
### Motivation:

The PR https://github.com/swiftlang/swift-org-website/pull/1347 reindended all the YAML files and that broke the download.swift.org OpenAPI doc.

### Modifications:

Fixed the indentation.

### Result:

The OpenAPI doc can be used for code generation again.


heckj (approved):

czechboy0:
Yeah we should investigate that separately, I just wanted to unbork it quickly so didn't dig into it yet

PR #1350, opened 2026-03-10
[Windows Install] Improve manual installation instructions

carlpeto:
Some users have found the installer instructions incomplete or misleading. This PR attempts to improve this.

As referenced in this thread: https://forums.swift.org/t/swift-not-found/80525/5

rdar://172184661

### DIFF
--- a/install/windows/_manual.md
+++ b/install/windows/_manual.md
@@ -2,19 +2,21 @@
 
 Swift depends on a number of developer tools when running on Windows, including the C++ toolchain and the Windows SDK. On Windows, typically these components are installed with [Visual Studio](https://visualstudio.microsoft.com). 
 
-The following Visual Studio components should be installed:
+The following Visual Studio components must be installed:
 
 | Name | Component ID |
 |-----------|------------------|
 | MSVC v143 - VS 2022 C++ x64/x86 build tools (Latest)[^1] | Microsoft.VisualStudio.Component.VC.Tools.x86.x64 |
 | MSVC v143 - VS 2022 C++ ARM64/ARM64EC build tools (Latest)[^1] | Microsoft.VisualStudio.Component.VC.Tools.ARM64 |
 | Windows 11 SDK (10.0.22000.0)[^2] | Microsoft.VisualStudio.Component.Windows11SDK.22000 |
+| C++ CMake tools for Windows | Microsoft.VisualStudio.Component.VC.CMake.Project |
+
 
 [^1]: At minimum, Swift requires the build tools that match your machine architecture. Installing other architectures is recommended in order to cross-compile Swift binaries to run on different machine architectures.
 
 [^2]: You may install a newer SDK instead.
 
-You should also install the following dependencies:
+You must also install the following dependencies:
 
 - [Python 3.10._x_](https://www.python.org/downloads/windows/) [^3]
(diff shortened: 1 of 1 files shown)


shahmishal (approved):

shahmishal:
cc: @compnerd

install/windows/_manual.md
```diff
@@ -2,19 +2,21 @@
...
 
-The following Visual Studio components should be installed:
+The following Visual Studio components must be installed:
 
 | Name | Component ID |
 |-----------|------------------|
 | MSVC v143 - VS 2022 C++ x64/x86 build tools (Latest)[^1] | Microsoft.VisualStudio.Component.VC.Tools.x86.x64 |
 | MSVC v143 - VS 2022 C++ ARM64/ARM64EC build tools (Latest)[^1] | Microsoft.VisualStudio.Component.VC.Tools.ARM64 |
 | Windows 11 SDK (10.0.22000.0)[^2] | Microsoft.VisualStudio.Component.Windows11SDK.22000 |
+| C++ CMake tools for Windows | Microsoft.VisualStudio.Component.VC.CMake.Project |
```
compnerd, line 12: Why are we adding the CMake tools? That installs CMake and Ninja. While those are useful for building some projects, I think that for the vast majority this shouldn't be needed.

install/windows/_manual.md
```diff
@@ -2,19 +2,21 @@
...
+
 
 [^1]: At minimum, Swift requires the build tools that match your machine architecture. Installing other architectures is recommended in order to cross-compile Swift binaries to run on different machine architectures.
 
 [^2]: You may install a newer SDK instead.
 
-You should also install the following dependencies:
+You must also install the following dependencies:
```
compnerd, line 19: Python is now being packaged as a component in the installer as of 6.2.4. If we add cmake above, I don't think that we need git - `git` is only a dependency for SPM.

install/windows/_manual.md
```diff
@@ -2,19 +2,21 @@
 
 Swift depends on a number of developer tools when running on Windows, including the C++ toolchain and the Windows SDK. On Windows, typically these components are installed with [Visual Studio](https://visualstudio.microsoft.com). 
 
-The following Visual Studio components should be installed:
+The following Visual Studio components must be installed:
```
compnerd, line 5: I think that `should` is still better - with embedded Swift, it feels like the dependencies here should be growing smaller for some set of folks.

compnerd (commented):
I think that the original text is much better than the suggested changes here.

`must` means that there is no wiggle room, where as with `should` indicates that there might be particular reason to deviate. (See RFC 2119/8174).

There are times where the user might need to deviate from these recommendations, so I would rather that we retain the current "should" over "must".

For the other dependencies, I think that there is the question of what is the motivation there.

install/windows/_manual.md
carlpeto: Valid points. We are stopping a bit to have a re-think on this.

install/windows/_manual.md
carlpeto: As an ex-embedded guy, I agree strongly with the sentiment. :) we are looking at other possible fixes.

carlpeto:
I think we went a different way on this. Closing PR.

PR #1375, opened 2026-03-25
Bring back more concrete Swift SDK for Android 6.3 install instructions

finagolfin:
#1371 simply copy-pasted the same verbose and non-specific instructions that the static linux SDK uses to this Android page, and removed the specific steps Marc had put together before, so bring that back, while still integrating the automated instructions Mishal asked for.

### DIFF
--- a/documentation/articles/swift-sdk-for-android-getting-started.md
+++ b/documentation/articles/swift-sdk-for-android-getting-started.md
@@ -17,63 +17,6 @@ To cross-compile Swift code for Android, you need three separate components:
 2. **The Swift SDK for Android**: the set of libraries, headers, and other resources needed to generate and run Swift code for the Android target.
 3. **The Android NDK**: the "Native Development Kit" for Android includes cross-compilation tools like `clang` and `lld` that are used by the host toolchain to cross-compile and link.
 
-### Installing the SDK
-
-#### (1) Prerequisites
-
-Before starting, please note the following requirements:
-
-* You need to install an open-source [Swift toolchain](/install/) from
-  swift.org.
-
-* If you are using macOS, note that you cannot use the toolchain
-  provided with Xcode to build programs using the SDK. Instead, you
-  must use the Swift compiler from the open-source toolchain (see
-  above).
-
-
-#### (2) Pre-Installation Notes
-
-Please be aware of:
-
-* Version compatibility: The Swift toolchain must match the version of
-  the Swift SDK for Android that you install.
-
-* Clean installation: If you previously installed an SDK for a
-  different Swift toolchain version, remove the old one before
-  installing the new one (see management commands below).
-
-* Checksum verification: When installing Swift SDKs from remote URLs,
-  you must pass a `--checksum` option with the corresponding checksum
-  provided by the SDK author.
-
-* Command pattern: The installation follows the pattern described in
-  the next sections.
-
-
-#### (3) Download and Install the Swift SDK for Android
-
-To obtain the Swift SDK for Android:
-
-* Visit the swift.org [installation
-  page](https://www.swift.org/install) for complete Swift SDK for Android
-  installation instructions, where you can download directly or click
-  "Copy install command".
-
-* For previous releases, navigate to "Previous Releases" on the
-  installation page.
-
-#### (4) Installation Commands Pattern
-
-The basic installation command follows this pattern:
-
-```console
-$ swift sdk install <URL-or-filename-here> [--checksum <checksum-for-archive-URL>]
-```
-
-You can provide either a URL (with corresponding checksum) or a local
-filename where the SDK can be found.
-
 <!--
 {% assign platform = site.data.builds.swift_releases.last.platforms | where: 'name', 'Android SDK'| first %}
 {% assign tag = site.data.builds.swift_releases.last.tag %}
@@ -84,32 +27,51 @@ filename where the SDK can be found.
 {% comment %} Generate branch information - ONLY major.minor {% endcomment %}
 -->
 
-For example, if you have installed the {{ tag }} toolchain, you would enter:
+#### 1. Install the Host Toolchain
+
+While `swift` may already be installed on your system (such as through an Xcode installation on macOS), using a cross-compilation Swift SDK requires using an open-source toolchain and that the Swift SDK version matches exactly.
+
+The easiest and recommended way to manage host toolchains on macOS and Linux is to use [the swiftly command](/swiftly/documentation/swiftly/getting-started) command. Once that has been setup, you can install the host toolchain with:
+
+```console
+$ swiftly install latest
+Fetching the latest stable Swift release...
+Installing Swift 6.3
+
+Installing package in user home directory...
+Swift 6.3 is installed successfully!
+$ swiftly use latest
+The global default toolchain has been set to `Swift 6.3`
+
+$ swift -version
+Apple Swift version 6.3 ({{tag}})
+Target: arm64-apple-macosx26.0
+Build config: +assertions
+```
+You can also find [direct links to the open-source Swift toolchains](/install/), if you prefer manually installing then adding Swift to your `PATH`.
+
+#### 2. Install the Swift SDK for Android
+
+Next, download and install the Swift SDK bundle using the `swift sdk` command:
 
 ```console
 $ {{ command }}
 ```
 
-This will download and install the corresponding Swift SDK for Android on
-your system.
-
-#### (5) Managing Installed SDKs
-
-After installation, you can manage your SDKs using these commands:
-
-List all installed SDKs:
+You should now see the Android Swift SDK included in the `swift sdk list` command:
 
 ```console
 $ swift sdk list
+{{tag}}_android
 ```
 
-Remove an SDK:
+Make sure to remove any old Android SDKs you have installed:
 
 ```console
 $ swift sdk remove <name-of-SDK>
 ```
 
-#### (6) Install and configure the Android NDK
+#### 3. Install and configure the Android NDK
 
 The Swift SDK for Android depends on the Android NDK, LTS version 27d or later, to provide the headers and tools necessary for cross-compiling to Android architectures. There are a variety of ways to [install the Android NDK](https://developer.android.com/ndk/guides), but the simplest is to just download and unzip the archive from the [NDK Downloads page](https://developer.android.com/ndk/downloads/#lts-downloads) directly.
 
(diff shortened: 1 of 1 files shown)


documentation/articles/swift-sdk-for-android-getting-started.md
```diff
@@ -202,7 +164,7 @@ Hello, world!
 
 Congratulations, you have built and run your first Swift program on Android!
 
-Android applications are typically not deployed as command-line executable tools. Rather, they are assembled into an `.apk` archive and launched as an app from the home screen. To support this, Swift modules can be built as shared libraries for each supported architecture and included in an app archive. Swift code can then be accessed from the Android app — which is typically written in Java or Kotlin — through [the Java Native Interface (JNI)](https://developer.android.com/training/articles/perf-jni) by using the low-level [swiftlang/swift-java-jni-core](https://github.com/swiftlang/swift-java-jni-core) and higher-level [swiftlang/swift-java](https://github.com/swiftlang/swift-java) interoperability library and tools.
```
finagolfin, line 205: Why link swift-java-jni-core here, when it says most won't want to use it?

documentation/articles/swift-sdk-for-android-getting-started.md
```diff
@@ -17,99 +17,64 @@ To cross-compile Swift code for Android, you need three separate components:
 2. **The Swift SDK for Android**: the set of libraries, headers, and other resources needed to generate and run Swift code for the Android target.
```
finagolfin, line 17: @timsneath, [you mentioned changing this earlier](https://github.com/swiftlang/swift-org-website/pull/1358#discussion_r2984521862), please let us know how it might be improved.

documentation/articles/swift-sdk-for-android-getting-started.md
```diff
@@ -17,99 +17,64 @@ To cross-compile Swift code for Android, you need three separate components:
...
 $ swift sdk list
+{{tag}}_android
 ```
-
-Remove an SDK:
+Make sure to remove any old Android SDKs you have installed:
 
 ```console
-$ swift sdk remove <name-of-SDK>
+$ swift sdk remove {{sdk_snapshot_tag}}_android
```
marcprux: This mostly affects people who had installed the preview snapshot builds, right? It might add unnecessary complexity to include this for the common case of a developer who hasn't been installing nightlies of the past months. If it isn't affecting a common case, I think we would be better just sticking with `swift sdk remove <name-of-SDK>`.

documentation/articles/swift-sdk-for-android-getting-started.md
```diff
@@ -17,99 +17,64 @@ To cross-compile Swift code for Android, you need three separate components:
...
-List all installed SDKs:
+You should now see the Android Swift SDK included in the `swift sdk list` command:
 
 ```console
 $ swift sdk list
+{{tag}}_android
 ```
-
-Remove an SDK:
+Make sure to remove any old Android SDKs you have installed:
```
marcprux, line 68: Is this to quiesce the confusing warnings about ambiguous triples? Or is there some concrete reason they need to "Make sure" to remove the old SDKs?

documentation/articles/swift-sdk-for-android-getting-started.md
marcprux: I think it is useful to call it out as a possible integration path, e.g. for apps that don't use much of the Java Android API but still need to have access to the jobject for the Android Context (e.g., a pure NDK game app). For example, https://github.com/swift-android-sdk/swift-android-native/blob/main/Package.swift needs "just a little" JNI, and so only relies on `swiftlang/swift-java-jni-core`.

documentation/articles/swift-sdk-for-android-getting-started.md
```diff
@@ -17,99 +17,64 @@ To cross-compile Swift code for Android, you need three separate components:
...
 ```console
-$ {{ command }}
+$ swiftly install latest
+Fetching the latest stable Swift release...
+Installing Swift {{version}}
+
+Installing package in user home directory...
+Swift {{version}} is installed successfully!
+$ swiftly use latest
+The global default toolchain has been set to `Swift {{version}}` (was 6.2.4)
```
marcprux: Can we just trim off the "(was 6.2.4)" so this ages well?

marcprux (changes requested):
A few minor nits from me, but overall I agree that this format was a bit nicer and simpler than the [Static Linux install instructions](https://www.swift.org/documentation/articles/static-linux-getting-started.html) that I had (hastily) copy/pasted for the 6.3 update.

I'll disclaim that I haven't yet run jekyll locally to preview this PR, so I can't vouch for the formatting. But I do think that the content is an improvement.

documentation/articles/swift-sdk-for-android-getting-started.md
finagolfin: Mind leaving it out for now and we can revisit later?

documentation/articles/swift-sdk-for-android-getting-started.md
finagolfin: It is always better to be more concrete, and adding this command is one of the few changes from the copy-paste I agree with, so I simply swapped in this snapshot tag.

Since this command only helps those who installed nightlies right now, I'd rather remove the command altogether if you are worried about that.

documentation/articles/swift-sdk-for-android-getting-started.md
finagolfin: Yep, that's it, we need to fix that bug in SwiftPM.

documentation/articles/swift-sdk-for-android-getting-started.md
finagolfin: I copy-pasted this from my install of a different release, would rather match what it does. We can fix the automated updating of this version later.

marcprux: I think it is better to trim it off. A fresh install of `swiftly use latest` wouldn't show the "was <old version>".

documentation/articles/swift-sdk-for-android-getting-started.md
marcprux: I'd also be OK with removing it altogether, but I do think it has some value to mention that old SDKs can be removed. I'm happy with whatever you decide, as long as the page doesn't contain references to `swift-DEVELOPMENT-SNAPSHOT-…` stuff that the common developer might find confusing.

documentation/articles/swift-sdk-for-android-getting-started.md
marcprux: I would prefer to leave it in.

documentation/articles/swift-sdk-for-android-getting-started.md
finagolfin: OK, will remove it.

Let me know what you think of the above matters and I will make all needed updates at one time.

documentation/articles/swift-sdk-for-android-getting-started.md
finagolfin: Why? You say below you don't even want to mention the nightlies, so why mention this even more niche library?

documentation/articles/swift-sdk-for-android-getting-started.md
finagolfin: Since this command currently only has value to remove the nightlies, I say we keep it like this, as that is a bug worth avoiding.

We can change it on the patch release.

documentation/articles/swift-sdk-for-android-getting-started.md
marcprux: I already said why. I think it is valuable information for different use cases.

finagolfin: Right, but my point is that these libraries are even more niche than the nightly snapshots, that point has not been addressed.

documentation/articles/swift-sdk-for-android-getting-started.md
timsneath: Thanks for flagging! How about I submit a separate PR for this paragraph specifically, so that you can get the rest of this out the door?

documentation/articles/swift-sdk-for-android-getting-started.md
marcprux: As I said, I believe it provides value and demonstrates that there is flexibility for integrating with JNI and Java on Android. Between https://github.com/swiftlang/swift-android-examples, https://github.com/swiftlang/swift-java-jni-core, and https://github.com/swiftlang/swift-java it shows that we have a rich set of support libraries and examples. I don't see any harm in leaving it in, especially as future projects (like the aforementioned https://github.com/swift-android-sdk/swift-android-native) will be building on top of it.

I'm not sure what other argument or evidence I can offer. I think it should stay.

marcprux: I'll also add that this mirrors the explicit mention of _both_ Swift Java and Swift Java JNI Core in the blog post (https://www.swift.org/blog/swift-6.3-released/#android), which was explicitly added by @shahmishal in https://github.com/swiftlang/swift-org-website/pull/1360#discussion_r2978868141.

finagolfin: The nightly snapshots also have value to a minority of users for testing, so it's not a question of value, but for how many people? And [your own README answers that question](https://github.com/swiftlang/swift-java-jni-core?tab=readme-ov-file#swift-java-interoperability) with:
"Most developers should prever the use of that package, as it features bullet proof ways to make your inter-op code safe and performant, by using either source generation (swift-java) when calling Swift from Java, or Swift macros (@JavaMethod, ...) when calling Java from Swift. Multiple safeguards and optimizations are built into the SwiftJava library and we highly recommend using it instead of this low level "raw JNI" package."

So based on that niche audience for this jni-core repo, I think it was a mistake for this to ever have been added here, which I didn't approve, so I'm removing it.

@timsneath or @shahmishal, Marc and I are not going to agree on this it seems, I ask that you merge this pull, and he can always submit another pull making a case for this link.

davelester: It'd be great to align this with what we've announced in the blog post, including `swift-java-jni-core` makes sense to me.

@finagolfin would it address your concern if we were to restructure the sentence so `swift-java` is mentioned first? Perhaps this moreso a question of what's emphasized.

finagolfin: It's not a question of emphasis: as the above quote makes clear, swift-java is aimed at developers writing apps, whereas swift-java-jni-core is aimed at a small segment of implementors, people who want direct JNI access.

For the same reason you didn't want the sentences about installing nightly snapshots in this article, this jni-core audience is even tinier. I was surprised it was mentioned in the blog post, but even more people will read this doc going forward.

I think it's best to just take out jni-core, but I'll do whatever the website workgroup wants.

davelester: On a related note, I sent [a PR](https://github.com/swiftlang/swift-java-jni-core/pull/15) to the `swift-java-jni-core` README to improve its description.

finagolfin: Ok, just let me know what you decide here too.

documentation/articles/swift-sdk-for-android-getting-started.md
```diff
@@ -202,7 +164,7 @@ Hello, world!
...
+Android applications are typically not deployed as command-line executable tools. Rather, they are assembled into an `.apk` archive and launched as an app from the home screen. To support this, Swift modules can be built as shared libraries for each supported architecture and included in an app archive. Swift code can then be accessed from the Android app — which is typically written in Java or Kotlin — through [the Java Native Interface (JNI)](https://developer.android.com/training/articles/perf-jni) by using [the swift-java interoperability library and tools](https://github.com/swiftlang/swift-java).
```
davelester:
```suggestion
Android applications are typically not deployed as command-line executable tools. Rather, they are assembled into an `.apk` archive and launched as an app from the home screen. To support this, Swift modules can be built as shared libraries for each supported architecture and included in an app archive. Swift code can then be accessed from the Android app — which is typically written in Java or Kotlin — through [the swift-java interoperability library and tools](https://github.com/swiftlang/swift-java), which handle [the Java Native Interface (JNI)](https://developer.android.com/training/articles/perf-jni) for you. For advanced uses, [Swift Java JNI Core](https://github.com/swiftlang/swift-java-jni-core) is also available as a low-level interface.
```

davelester: @marcprux @finagolfin would this revision strike a balance you both are OK moving forward with?

marcprux: LGTM!

davelester (approved):

documentation/articles/swift-sdk-for-android-getting-started.md
ktoso: Thanks folks the PR as-was-merged looks OK. Please ping me whenever swift-java is brought up, I almost missed this.

We do _not_ want developers to use swift-java-jni-core, and I don't think we want to promote it at all. It exists solely as a compromise to make some folks happier to share their work, but ultimately swift-java is the interop package and the way we're promoting.

Thank you for the PR Dave, will get that merged

PR #1396, opened 2026-04-21
GHA: Add PR dependency check

bkhouri:
Add a workflow that will check if dependent PRs/Issues are merged. If not, fail the build.

### DIFF
--- a/.github/workflows/pull_request_dependency_check.yaml
+++ b/.github/workflows/pull_request_dependency_check.yaml
@@ -1,0 +1,13 @@
+name: "Dependencies"
+
+on:
+  pull_request_target:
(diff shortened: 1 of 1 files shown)


.github/workflows/pull_request_dependency_check.yaml
```diff
@@ -0,0 +1,13 @@
+name: "Dependencies"
```
bkhouri, line 1: FYI. this same workflow exists in swift-package-manager.

https://github.com/swiftlang/swift-package-manager/blob/main/.github/workflows/pull_request_dependency_check.yaml

bkhouri:
Cleaning up my open PR. Closing.

Issue #1388, opened 2026-04-07
Broken links (404) on | Getting Started with Swift SDKs for WebAssembly


zaneenders:
## Describe the bug

Getting a 404 for the following links at the bottom of [Getting Started with Swift SDKs for WebAssembly](https://www.swift.org/documentation/articles/wasm-getting-started.html).

[Zero to Swift with Neovim](https://www.swift.org/documentation/articles/zero-to-swift-nvim/)
[Zero to Swift with Emacs](https://www.swift.org/documentation/articles/zero-to-swift-emacs/)

## To Reproduce

- Open [Getting Started with Swift SDKs for WebAssembly](https://www.swift.org/documentation/articles/wasm-getting-started.html).

- Scroll to bottom of the page. 

- Click either of these links 

[Zero to Swift with Neovim](https://www.swift.org/documentation/articles/zero-to-swift-nvim/)
[Zero to Swift with Emacs](https://www.swift.org/documentation/articles/zero-to-swift-emacs/)

## Expected behavior

Link to article.

## Actual behavior

404 page.

### Desktop (please complete the following information):

Safari Version 26.4 (21624.1.16.11.4)

kuhl:
This issue is now fixed. You may close it.

zaneenders:
Thank you!

PR #1398, opened 2026-04-27
Fix Wasm Getting Started broken links for issue #1388

kuhl:
URLs on page were directories and not the .html files. Documentation was returning 404 errors when user clicks links.

### Motivation:

Fulfills issue #1388 where links were directories and not .html links.

### Modifications:

Changed the directory links to .html links

### Result:

No more 404 error.

### DIFF
--- a/documentation/articles/wasm-getting-started.md
+++ b/documentation/articles/wasm-getting-started.md
@@ -124,7 +124,7 @@ This section shows you how to configure your development environment for Swift W
 
 ### Visual Studio Code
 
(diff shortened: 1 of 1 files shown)


heckj (approved):
Simpler would be to just remove the trailing `/`, which routes correctly, but this works too.

shahmishal (approved):

kuhl:
> Simpler would be to just remove the trailing `/`, which routes correctly, but this works too.

I just tested this and that doesn't work.

PR #1400, opened 2026-04-28
Android getting started: try to finally fix the name of the previous SDK

finagolfin:
@federicobucchi, please take a look, I see you used `slice` in the `index.md`, as I'm a Jekyll noob.

### DIFF
--- a/documentation/articles/swift-sdk-for-android-getting-started.md
+++ b/documentation/articles/swift-sdk-for-android-getting-started.md
@@ -30,8 +30,9 @@ Cross-compilation for Android requires installing three separate components:
 {% assign platform = site.data.builds.swift_releases.last.platforms | where: 'name', 'Android SDK'| first %}
 {% assign tag = site.data.builds.swift_releases.last.tag %}
 {% assign version = site.data.builds.swift_releases.last.name %}
-{% assign previous_releases = site.data.builds.swift_releases | offset:1 %}
-{% assign previous_sdk_name = previous_releases.last.name | append: "_android" %}
+{% assign release_count = site.data.builds.swift_releases.size %}
+{% assign previous_release = site.data.builds.swift_releases | slice: release_count - 2 %}
+{% assign previous_sdk_name = previous_release.tag | append: "_android" %}
 {% assign tag_downcase = site.data.builds.swift_releases.last.tag | downcase %}
 {% assign base_url = "https://download.swift.org/" | append: tag_downcase | append: "/android-sdk/" | append: tag | append: "/" | append: tag %}
 {% assign command = "swift sdk install " | append: base_url | append: "_android" | append: ".artifactbundle.tar.gz --checksum " | append: platform.checksum %}


documentation/articles/swift-sdk-for-android-getting-started.md
```diff
@@ -30,8 +30,9 @@ Cross-compilation for Android requires installing three separate components:
 {% assign platform = site.data.builds.swift_releases.last.platforms | where: 'name', 'Android SDK'| first %}
 {% assign tag = site.data.builds.swift_releases.last.tag %}
 {% assign version = site.data.builds.swift_releases.last.name %}
-{% assign previous_releases = site.data.builds.swift_releases | offset:1 %}
-{% assign previous_sdk_name = previous_releases.last.name | append: "_android" %}
+{% assign release_count = site.data.builds.swift_releases.size %}
+{% assign previous_release = site.data.builds.swift_releases | slice: release_count - 2 %}
```
finagolfin, line 34: 2 because these arrays are zero-indexed.

documentation/articles/swift-sdk-for-android-getting-started.md
shahmishal, line 84: Why do we need to uninstall the previous SDK? Is it because the `swift sdk` doesn’t recognize the toolchain associated with the installed SDK?

finagolfin: Yep, exactly, it will fail using the `build` command we give if both versions of the Android SDK are installed. I want to get that working better by submitting this SwiftPM fix for the next patch release, swiftlang/swift-package-manager#9955, after which we can tell people to use that full command instead and remove this caution.

For the next 6.4 release, we should get [SDK version checking working](https://forums.swift.org/t/stricter-searching-for-swift-sdks-when-building/86011), which would be the easiest to use, but will require more work on our end to set that up.

finagolfin:
Hmm, Federico not too active here lately, I think this will finally work.

finagolfin:
@shahmishal, did I finally get this working? 🥲

documentation/articles/swift-sdk-for-android-getting-started.md
federicobucchi: Good catch on the zero-indexing.
`slice` returns an array here, not a single release object, so I am not sure if `previous_release.name` would work as expected.
Can you try:

```liquid
{% assign release_count = site.data.builds.swift_releases.size %}
{% assign previous_release = site.data.builds.swift_releases[release_count | minus: 2] %}
{% assign previous_sdk_name = previous_release.name | append: "_android" %}
```

finagolfin: > slice returns an array here, not a single release object, so I am not sure if previous_release.name would work as expected.

Are you sure? [The doc indicates otherwise when a single number is specified](https://shopify.github.io/liquid/filters/slice/), like I do here, but I have not tested it locally.

If you have a local setup you can try this out on, that would be great. 👍

federicobucchi: You may be right — I was thinking of the general array slicing behavior. Looking at the Liquid docs again, it does seem that passing a single index may return the element directly rather than a sub-array.
I have not tested it locally yet, so I cannot say with certainty.
I didn't test it locally for time constraints, but I would highly recommend to test changes locally

finagolfin:
@shahmishal, I think we can get this in, should work. If not, I'll finally get it setup locally, which I've been trying to avoid for this single foray into Liquid templating. 😉

finagolfin:
Ping, this should finally work. At worst, it is already wrong in the tutorial now and will stay wrong. 😉

shahmishal (approved):

PR #1403, opened 2026-05-04
[cxx-interop] Document support for C++ stdlib types

j-hui:
### Motivation:

This list has not been kept up to date as we've added more support for C++ interop.

### Modifications:

Document which C++ stdlib types we support, especially those that we have tests for, or special compiler handling for.

### Result:

The list of supported types.


j-hui:
CC: @egorzhdan @Xazax-hun @susmonteiro @hnrklssn @patrykstefanski @ravikandhadai

Xazax-hun (commented):
The change looks good to me, but I almost wonder if we reached a point where we should only mention what is NOT supported because most things (even new containers) should just work out of the box.

j-hui:
> I almost wonder if we reached a point where we should only mention what is NOT supported because most things (even new containers) should just work out of the box.

I was thinking that too, but I think we should also expand our documentation about what we _do_ support and have examples using each of them

documentation/cxx-interop/status/index.md
```diff
@@ -189,18 +189,17 @@ type of such data member is supported in Swift.
...
-- Specializations of `std::set`, `std::multiset` and `std::unordered_set`
-- Specializations of `std::optional`
-- Specializations of `std::shared_ptr`
-- Specializations of `std::unique_ptr`
-- Specializations of `std::array`
-
-Other standard library types, like `std::function` and `std::variant`, are not
-yet supported in Swift.
+- `std::string`, `std::u16string`, `std::u32string`, `std::string_view`
+- Specializations of `std::pair` and `std::optional`
```
egorzhdan, line 193: Minor: did you group `std::pair` and `std::optional` together for a reason?

j-hui: Yeah, I did so here e4f57d61663a30ded9f9fe4f39db894f568b32ef since they're both small type-level combinators for structuring values

documentation/cxx-interop/status/index.md
```diff
@@ -189,18 +189,17 @@ type of such data member is supported in Swift.
...
+- Specializations of `std::vector`, `std::array`, and `std::span`
+- Specializations of `std::map`, `std::unordered_map` and `std::multimap`
+- Specializations of `std::set`, `std::unordered_set` and `std::multiset`
+- Specializations of `std::shared_ptr` and `std::unique_ptr`
```
susmonteiro: Do we want to mention that `std::unique_ptr` is not fully supported in Windows?

susmonteiro:
> The change looks good to me, but I almost wonder if we reached a point where we should only mention what is NOT supported because most things (even new containers) should just work out of the box.

One example is `std::list`. We don't have an overlay protocol for it, but you can kind of use it in Swift. However, perhaps we should wait for Swift 6.4 to include it in the list, because with the conformance to `BorrowingSequence` we'll be able to iterate them.

j-hui:
> One example is std::list. We don't have an overlay protocol for it, but you can kind of use it in Swift.

This is true, but the existence of an overlay protocol implies we guarantee a stronger level support than what you'd get out of the box + those protocols should also mean we are actively testing the support in our regression tests. For now, I think we should keep it to known good cases until we've thoroughly evaluated how well interop works for non-overlay types + gotten to a point where we can actually enumerate what _doesn't_ work.

susmonteiro (approved):
Nice, thanks!

documentation/cxx-interop/status/index.md
j-hui: @egorzhdan do you feel strongly about separating these types? I'm happy to go either way 🙂

egorzhdan: I don't feel strongly about this!

j-hui:
@heckj This PR is ready to be merged, could you please take a look? Thanks!

heckj (approved):

PR #1407, opened 2026-05-07
fix: Add "Copy to Clipboard" button to markdown code blocks (#1340)

### LINKED ISSUE #1340: Missing Copy Button on hover
## Describe the bug

There should be a button to copy the command from the code block in Linux installation instruction which makes it easier to copy

Image

## To Reproduce

1. Go to (https://www.swift.org/install/linux/swiftly/)
2. Hover on code blocks

## Expected behavior

it is aspected that on hover there should be a copy button 

Image

## Actual behavior

there is no copy button on hover

### Desktop (please complete the following information):

- OS: Ubuntu 24
- Browser chrome

ShameelMohamed:
**fix: Add "Copy to Clipboard" button to markdown code blocks (#1340)**

### Motivation:

Code blocks in the documentation (such as the Linux installation guide) currently require manual text selection to copy terminal commands. Issue #1340 requested a copy button on hover to improve the developer experience and make copying commands seamless and error-free.

### Modifications:

- **CSS (`_syntax.scss`):** Added styling for a `.copy-button`. It remains hidden (`opacity: 0`) by default and transitions to visible only when the parent `<pre>` block is hovered, maintaining the site's clean aesthetic. Uses Apple-standard styling (pill shape, backdrop blur).
- **JavaScript (`application.js`):** Added an initialization function that injects a copy button into all `<pre>` blocks (excluding those with a `.no-copy` class). The logic clones the target node before extraction to ensure the button's own text ("Copy") is never accidentally copied.
- **Terminal Output Filtering:** Implemented a line-boundary Regex (`/\n\s*--\s*(?:\n|$)/`) to intelligently detect the site's `--` output separators. This safely strips terminal output from the clipboard while preventing accidental truncation on standard CLI flags (e.g., `swift --version` or `ls --all`).
- **Clipboard API:** Implemented `navigator.clipboard.writeText` to handle the copy action, wrapping it in a `try/catch` block for safety, and including a temporary "Copied!" visual state to provide immediate user feedback.

### Result:

Users hovering over any code block will now see a clean copy button in the top right corner. Clicking it successfully copies the code to their clipboard (automatically stripping out any trailing terminal output) and shows a brief confirmation state. 

Closes #1340

### DIFF
--- a/assets/javascripts/application.js
+++ b/assets/javascripts/application.js
@@ -88,23 +88,53 @@ layout: source
       })
     }
 
-    if (navigator && navigator.clipboard) {
-      const copyCodeButtons = document.querySelectorAll('[class^="language-"] button');
+    function initializeCopyButtons() {
+      if (!(navigator && navigator.clipboard)) {
+        return;
+      }
 
-      copyCodeButtons.forEach(function(button) {
-        const codeElement = button.parentElement.querySelector('code');
-        const originalText = button.innerText;
+      const copyIcon = '<svg xmlns="http://www.w3.org/2000/svg" viewBox="0 0 24 24" width="24" height="24" fill="currentColor"><path d="M16 1H4c-1.1 0-2 .9-2 2v14h2V3h12V1zm3 4H8c-1.1 0-2 .9-2 2v14c0 1.1.9 2 2 2h11c1.1 0 2-.9 2-2V7c0-1.1-.9-2-2-2zm0 16H8V7h11v14z"/></svg>';
+      const copiedIcon = '<svg xmlns="http://www.w3.org/2000/svg" viewBox="0 0 24 24" width="24" height="24" fill="currentColor"><path d="M9 16.2L4.8 12l-1.4 1.4L9 19 21 7l-1.4-1.4L9 16.2z"/></svg>';
 
-        button.addEventListener('mousedown', async function() {
-          await navigator.clipboard.writeText(codeElement.innerText);
+      document.querySelectorAll('pre:not(.no-copy)').forEach(function(pre) {
+        if (pre.querySelector('.copy-button')) {
+          return;
+        }
 
-          button.innerText = 'Copied!';
+        const codeElement = pre.querySelector('code') || pre;
+        const button = document.createElement('button');
 
-          setTimeout(() => {
-            button.innerText = originalText;
-          }, 1000);
+        button.type = 'button';
+        button.className = 'copy-button';
+        button.setAttribute('aria-label', 'Copy code');
+        button.title = 'Copy code';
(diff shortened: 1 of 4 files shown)


heckj:
Hey @ShameelMohamed  - thanks for the pull request. I'm not deeply familiar with the specifics of javascript, so I used some assistance to understand the PR, and wanted to ask about a few things:

First - this seems to attach to every <pre> site wide, including non-code - but the PR doesn't mention anything about making a pass through content to select what's enabled or not. Since you aimed for default-on (which may be valid), this'll have a much bigger impact. The previous mechanism in place (I think) was meant for each code block to be wired up *if* they wanted to be copied - with `.code-box pre code` - was there a limitation there you didn't think was correct, to where you're suggesting making it universal? That seems like it might be too big of a sweep.

Second -  there seems to be an accessibility regression for the button that shows up on focus because `display:none` is used, which removes it from access to keyboard/mouse interactions; there's some front-matter in Jekyll that's removed which causes a copyright header to no longer emit, and the clever -- regex ends up meaning that what you see on the screen won't be what you get copied the clipboard - useful for the install pages you were tackling, but the broad impact may be negative. It may make a lot more sense to re-work the install pages content so that you didn't need to do this "front matter exclusion" and still support the copy to code. What that an option you considered?

ShameelMohamed:
Hi @heckj ,

Thank you for the thorough review and for bringing up these excellent points! I completely agree with your assessment. It makes a lot of sense to keep the implementation localized and predictable.

Here is my plan to address each of your concerns in the next commit:

1. Scope of the Button (Universal vs. Opt-in)
You are absolutely right that targeting every pre tag site-wide is too aggressive. I will revert the selector to explicitly target .code-box pre code (or a specific opt-in class) so it only applies to designated code blocks.

2. Accessibility Regression
Great catch on the keyboard navigation. Using display: none was a mistake. I will update the CSS to use an opacity: 0 to opacity: 1 transition, combined with :focus and :focus-within states, ensuring the button remains fully accessible for keyboard tabbing and screen readers.

3. Missing Jekyll Front-Matter
That was an accidental deletion on my part while testing locally. I will restore the missing front-matter immediately so the copyright headers emit correctly.

4. The -- Regex and WYSIWYG
I completely agree with the "what you see is what you get" philosophy. Stripping output magically via regex introduces too much unpredictability. I will remove the regex logic entirely. Instead, as you suggested, I will update the markdown content on the installation pages to separate the terminal commands from the expected console outputs into distinct blocks.

I will get to work on these changes and push an updated commit shortly. Thanks again for the guidance!

heckj:
possibly related fix/alternate: https://github.com/swiftlang/swift-org-website/pull/1389

PR #1250, opened 2025-11-21
Refine Typography

jesseaborden:
### Motivation:

Start to unify typography styles into shared classes to cut down on variations. 

### Modifications:

- Created typography classes for titles and body copy. 
- Updated blog markup and styles with new classes. 
- Removed unused blog styles.
- Removed mobile blog styles
- Updated landing with new typography classes.

### Result:

Landing and blog pages are now using new typography classes.

PR #1429, opened 2026-06-04
Centers title of above featured blog posts.

davelester:
Simple one-line change, which would adjust the title so "Latest Swift News" is centered on [swift.org/blog](https://www.swift.org/blog/).

Screenshot after the change:
Screenshot 2026-06-04 at 15 55 52


davelester:
This was fixed as part of #1250.

PR #1435, opened 2026-06-09
[cxx-interop] Overhaul FRT documentation

j-hui:
This rewrite addresses some gaps and inaccuracies in the existing documentation for FRTs, and cleans up the flow and writing.

Also, add some missing annotations.

### DIFF
--- a/documentation/cxx-interop/index.md
+++ b/documentation/cxx-interop/index.md
@@ -1314,210 +1314,399 @@ takesVectorType(&vector) // 'vector' is not copied!
 
 ## Mapping C++ Types to Swift Reference Types
 
-The Swift compiler allows you to annotate some C++ types and import them as reference types (or `class` types) in Swift. Whether a C++ type should be imported as a reference type is a complex question, and there are two primary criteria that go into answering it.
+By default, Swift imports C++ `class` and `struct` types as Swift structures
+with [value semantics](#c-structures-and-classes-are-value-types-by-default).
+This works well for small, self-contained types that are typically copied by value.
+Some C++ types, however, are designed around identity and shared ownership rather
+than value semantics. For these types, you can use customization macros from the
+`<swift/bridging>` header to instruct Swift to import them as **foreign reference
+types**.
 
-The first criterion is whether object identity is part of the "value" of the type. Is comparing the address of two objects just asking whether they're stored at the same location, or is it deciding whether they represent the "same object" in a more significant sense?
+Foreign reference types are imported with reference semantics that resemble
+Swift classes: variables hold references to the same underlying object,
+modifications are visible through every reference, and a `let`-declared
+reference can still call non-`const` C++ methods. Pointers and references to a
+foreign reference type in C++ bridge directly to the Swift reference type,
+rather than to an `UnsafePointer` variant.
 
-The second criterion is whether objects of the C++ class are always passed around by reference.  Are objects predominantly passed around using a pointer or reference type, such as a raw pointer (`*`), C++ reference (`&` or `&&`), or a smart pointer (like `std::unique_ptr` or `std::shared_ptr`)?  When passed by raw pointer or reference, is there an expectation that that memory is stable and will continue to stay valid, or are receivers expected to copy the object if they need to keep the value alive independently?  If objects are generally allocated and remain at a stable address, even if that address is not semantically part of the "value" of an object, the class may be idiomatically a reference type. This will sometimes be a judgment call for the programmer.
+Types that are never stack-allocated and are predominantly passed by raw
+pointer (`*`), C++ reference (`&` or `&&`), or smart pointer (such as
+`std::unique_ptr` or `std::shared_ptr`) are natural candidates for being
+imported as reference types. Reference types also frequently delete the copy
+constructor to prevent accidental value copies that would bypass reference
+counting.
 
-The first and most important criteria is often not possible for a compiler to answer automatically by just looking at the code. If you want the Swift compiler to map a C++ type to a Swift reference type, you must annotate the C++ type with one of the following
-customization macros from the `<swift/bridging>` header:
+Swift supports three foreign reference type variants, declared by applying one
+of the following customization macros from `<swift/bridging>`:
 
-- [`SWIFT_IMMORTAL_REFERENCE`](#immortal-reference-types)
-- [`SWIFT_SHARED_REFERENCE`](#shared-reference-types)
-- [`SWIFT_UNSAFE_REFERENCE`](#unsafe-reference-types)
+- [`SWIFT_SHARED_REFERENCE`](#shared-reference-types) for reference-counted
+  types.
+- [`SWIFT_IMMORTAL_REFERENCE`](#immortal-reference-types) for persistent
+  objects like singletons.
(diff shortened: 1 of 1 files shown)


j-hui:
CC @egorzhdan @Xazax-hun @susmonteiro @patrykstefanski @hnrklssn @ravikandhadai

documentation/cxx-interop/index.md
```diff
@@ -1311,212 +1311,399 @@ takesVectorType(&vector) // 'vector' is not copied!
 
 ## Mapping C++ Types to Swift Reference Types
 
-The Swift compiler allows you to annotate some C++ types and import them as reference types (or `class` types) in Swift. Whether a C++ type should be imported as a reference type is a complex question, and there are two primary criteria that go into answering it.
+By default, Swift imports C++ class and structure types as Swift `struct` types
```
Xazax-hun: Nit: I wonder if we should not say structure types but rather `class`es and `struct`s, just referring to the keywords. I am not sure how often are we actually say structure types.

documentation/cxx-interop/index.md
```diff
@@ -1311,212 +1311,399 @@ takesVectorType(&vector) // 'vector' is not copied!
...
+with [value semantics](#c-structures-and-classes-are-value-types-by-default).
+This works well for types that are small, self-contained, and copied by value.
```
Xazax-hun: Nit: I wonder if you wanted to say copy has value semantics. I think I more often heard the copied by value terminology in the context of passing arguments to functions rather than the behavior of a type.

documentation/cxx-interop/index.md
```diff
@@ -1311,212 +1311,399 @@ takesVectorType(&vector) // 'vector' is not copied!
...
 
-#### Calling conventions when passing Shared Reference Types from Swift to C++
+#### Inheritance and Virtual Member Functions
 
-If a C++ shared reference type is passed as an argument to a C++ API from Swift, the Swift compiler guarantees that the passed value would be alive. 
-Swift also retains the ownership of the value.
-In other words, the argument is passed at `+0` and there is no transfer of ownership.
-The C++ function should not assume that it has the ownership of the value and should do necessary retain operations if it is needs to take ownership.
-The C++ function is responsible for ensuring that the value pointed to by the parameter is alive during and at the end of the function call.
+A C++ class that inherits from a foreign reference type is itself a foreign
```
Xazax-hun, line 1571: Should we say a unique foreign reference type? If there are multiple foreign reference bases or some other reason the inference does not work we might fall back to importing as value types? Or are we just not importing the type at all?

documentation/cxx-interop/index.md
```diff
@@ -1311,212 +1311,399 @@ takesVectorType(&vector) // 'vector' is not copied!
...
-var obj = SharedObject.create()
-receiveSharedObject(obj) // Swift guarantees that obj is alive and it is passed at +0
+let circle = Circle(5.0)
+let rect = Rectangle(3.0, 4.0)
+print(circle.radius()) // Prints "5.0".
+print(rect.width()) // Prints "3.0".
+print(rect.height()) // Prints "4.0".
 ```
 
+Virtual methods, including pure virtual methods, can be called on foreign
```
Xazax-hun, line 1612: Should we somehow mention the new feature of statically dispatching to a base method via `super`?

Xazax-hun (commented):
Overall, looks good, some questions/nits inline.

Should this mention `SWIFT_REFCOUNTED_PTR` somewhere?

Should this mention the new feature where single inheritance across shared references are reflected as inheritance in Swift (making upcasts work)?

documentation/cxx-interop/index.md
egorzhdan: That feature only exists in the development snapshots of Swift 6.5, it isn't available in a released compiler yet.

documentation/cxx-interop/index.md
j-hui: FWIW Swift seems to prefer the full spelling of "structure", e.g., [here](https://docs.swift.org/swift-book/documentation/the-swift-programming-language/classesandstructures) and in other chapters.

I think I will just flip this and say something like:

```suggestion
By default, Swift imports C++ `class` and `struct` types as Swift structures
```

documentation/cxx-interop/index.md
j-hui: Right now, we are falling back to importing a type that inherits from multiple FRT bases as a value (and emitting a diagnostic when it is used), but I don't think that's the right behavior. The fact that it inherits from _any_ FRT means that it is probably _supposed_ to have reference semantics, even if the annotations are such that we can't validly import it as such. Thus, we should import it as an (invalid) FRT and diagnose it as a reference type, rather than as a value type like we currently do.

As such, I prefer to leave this criterion deliberately loose here, because as I envision it, a C++ class that inherits from multiple FRTs is also itself a FRT, albeit an invalid one that cannot be used in Swift.

j-hui:
> Should this mention SWIFT_REFCOUNTED_PTR somewhere?

@Xazax-hun the `SWIFT_REFCOUNTED_PTR` now directly follows the writing here. Do you reckon that is sufficient?

documentation/cxx-interop/index.md
j-hui: I tweaked the phrasing slightly, but ended up settling back at "copied by value". It felt awkwardly circular to say "we import C++ structs as Swift structs with value semantics, which works well when those C++ structs have value semantics".

Xazax-hun:
Sounds good!

shahmishal:
@j-hui What is the status of this PR? Is it ready to be merged?

j-hui:
@shahmishal I'm still waiting on approving reviews from @egorzhdan and @Xazax-hun

Xazax-hun (approved):
LGTM!

j-hui:
I rebased on top of https://github.com/swiftlang/swift-org-website/pull/1495. This overhaul already had most of content from that PR so there wasn't anything to add.

There are two points missing from my writing:

- inherited annotations coming from API notes behave the same as inherited annotations from the header file. I'm not sure it's worth mentioning because in general the programming model for API notes should be that they are equivalent to annotating the header directly, so there's no point pointing out this specific case.
- different kinds of reference types cannot be mixed within an inheritance chain. Whether this is true depends on which inheritance chain is being referred to. It is possible to mix and match annotations along the same C++ inheritance chain; the annotation on the derived class just overrides that on the base. The effect is that these become independent Swift inheritance chains, so in Swift these reference type kinds cannot be mixed because such a chain cannot be formed in the first place.

@egorzhdan what do you think?

egorzhdan:
> I'm not sure it's worth mentioning because in general the programming model for API notes should be that they are equivalent to annotating the header directly

Fair enough, it should be fine to omit this.

> different kinds of reference types cannot be mixed within an inheritance chain

I think this statement it partially true assuming the inheritance relation is actually used in C++, e.g. there is a C++ function that has a return type of `Base*` but occasionally returns `Derived*`. Using different refcounting annotations in cases like this would lead to surprising behavior: Swift would use `Base`'s refcount operations even for `Derived*`.

j-hui:
@shahmishal this PR should be ready to merge!

j-hui:
@heckj could you please take a look at this? This PR is ready to merge. Thank you!

heckj (approved):

PR #1437, opened 2026-06-11
Add blogpost "Swift at Apple: Migrating the TrueType Hinting Interpreter"

numist:
### Motivation:

This year Holly shared that we rewrote the TrueType hinting interpreter in memory safe Swift. The new interpreter is also exhaustively tested, faster than the code it replaced, and we want to share that story with the world.

### Modifications:

Adds new blog post: "Swift at Apple: Migrating the TrueType Hinting Interpreter" with a publication date of 2026-06-11 16:00:00

PR #1439, opened 2026-06-12
Always provide full text RSS feed entries

keith:
Typically if a user is reading a blog in an RSS reader they want to be
able to read the entire post.

### DIFF
--- a/atom.xml
+++ b/atom.xml
@@ -23,7 +23,7 @@
     <link href="{{ site.url }}{{ post.url }}"/>
     <updated>{{ post.date | date_to_xmlschema }}</updated>
     <id>{{ site.url }}{{ post.url }}</id>
-    <content type="html">{% if post.excerpt_separator %}{{ post.excerpt | xml_escape }}{% else %}{{ post.content | xml_escape }}{% endif %}</content>
+    <content type="html">{{ post.content | xml_escape }}</content>
   </entry>
   {% endfor %}
 </feed>
--- a/atom.xml
+++ b/atom.xml
@@ -8,7 +8,7 @@
   <updated>{{ site.time | date_to_xmlschema }}</updated>
   <id>tag:swift.org,2015-12-03:Swift</id>
 
-  {% assign posts = site.posts | sort: 'updated' %}
+  {% assign posts = site.posts | sort: 'date' | reverse %}
   {% for post in posts limit: 20 %}
   <entry>
     <title>{{ post.title }}</title>


shahmishal:
This regressed with https://github.com/swiftlang/swift-org-website/pull/1437

heckj:
I'm seeing a warning with one of the other blog posts when I rebased and ran this over main:
```
Suggestion: if(sass($map): $map; else: $default)

More info: https://sass-lang.com/d/if-function

   ╷
41 │   @return if($map, $map, $default);
   │           ^^^^^^^^^^^^^^^^^^^^^^^^
   ╵
    /Users/joeheck/src/swift-org-website/assets/stylesheets/core/_maps.scss 41:11  @use
    /Users/joeheck/src/swift-org-website/assets/stylesheets/_core.scss 3:1         @use
    /Users/joeheck/src/swift-org-website/assets/stylesheets/application.scss 1:1   root stylesheet
    Liquid Warning: Liquid syntax error (line 27): Expected end_of_string but found dash in "{{site.data.builds.swift_releases | slice: release_count - 2 }}" in documentation/articles/swift-sdk-for-android-getting-started.md
```

unrelated to this revert, in the process of testing (or trying to)

davelester:
I believe the PR #1437 included JavaScript which broke RSS, so it adopted an excerpt. That shouldn't have removed it entirely, though.

The expected behavior here should be that posts that specify the `excerpt_separator` yaml metadata (currently that's limited to [Swift at Apple: Migrating the TrueType Hinting Interpreter](https://www.swift.org/blog/migrating-truetype-hinting-to-swift/) and [Swift OpenAPI Generator 1.0 Released](https://www.swift.org/blog/swift-openapi-generator-1.0/)) are excerpted, otherwise the full post should appear.

atom.xml
heckj:
```suggestion
  {% assign posts = site.posts | sort: 'date' | reverse %}
```

Fixing a long-standing ordering so that full content shows, but in reverse date order explicitly

(`site.posts` I believe is already sorted in reverse date order now, but it hasn't always been so - so including it makes things explicit)

heckj:
Functionally, this is working for most of the posts - but the truetype hinting interpreter post (added in #1437) won't appear here - the atom.xml relies having `updated` in the front-matter:
```
{% assign posts = site.posts | sort: 'updated' %}
```

With this change now inline, I've verified (dumping the raw atom.xml and using NetNewsWire) that I could read and review the posts. The SVG content in the truetype article doesn't flow through correctly - but I'm not sure I'd expect it to.

When I reverted the atom.xml, I noticed that the TrueType post _wasn't_ listed at the top of the feed, due to sorting on `updated`, which wasn't reflected properly - so I've applied an update to include it through.

keith:
> The expected behavior here should be that posts that specify the excerpt_separator yaml metadata (currently that's limited to [Swift at Apple: Migrating the TrueType Hinting Interpreter](https://www.swift.org/blog/migrating-truetype-hinting-to-swift/) and [Swift OpenAPI Generator 1.0 Released](https://www.swift.org/blog/swift-openapi-generator-1.0/)) are excerpted, otherwise the full post should appear.

this is what I was suggesting we change. I don't think RSS users ever want shortened articles and they can always decide to open the original if they'd like. My assumption about this type of excerpting is it is common for websites that depend on you visiting their actual pages for ad revenue (for example the verge does this unless you have a subscription) and it doesn't seem like that's something we should care about for this site, so i don't see any upside to this for RSS users?

heckj:
Sorry Keith - I should have included some of this detail up front - we chatted about this PR in the Website Workgroup meeting earlier today, and resolved that we do want to expose the whole post via RSS, and then went hunting back to sort things. When I was testing locally with this update, I noticed the post wasn't missing - so I extended the revert to deal with the incorrect sorting situation which has been in place for years, and broke somewhere where we missed it.

Thank you for prompting this repair and spotting the break. We'd prefer to have the whole content available via RSS/Atom

heckj (approved):

keith:
ah ok, thanks for the info!

davelester:
I believe some posts like the most recent TrueType post require that we excerpt them because they include JavaScript in the post, hence their use of the excerpt functionality. However that should not be the default behavior, only one overridden when necessary.

heckj:
@davelester the existing atom.xml has a means of doing that - took me half a minute to dig it out, but the gist is that if you set `excerpt_separator` in the front-matter for a specific post, then the Atom feed will only present the excerpt for _that post_

The example I found of this operating in earlier posts is [_posts/2024-01-31-swift-openapi-generator-1.0.md](https://github.com/swiftlang/swift-org-website/blob/main/_posts/2024-01-31-swift-openapi-generator-1.0.md), and it looks like that's set in the truetype post already

PR #1442, opened 2026-06-16
fixes a warning in the Android getting started Liquid 

heckj:
Article uses inline arithmetic- which is working, but throwing a warning. This updates is to use a liquid template mechanism as a replacement to achieve the same result.


shahmishal (approved):

finagolfin:
Still [not working](https://www.swift.org/documentation/articles/swift-sdk-for-android-getting-started.html)? I see this: `swift sdk remove _android`, instead of `remove swift-6.3.1-RELEASE_android`, as was my intent.

Issue #954, opened 2025-04-01
System package list for tar installation is out of date


cmcgee1024:
## Describe the bug

When going to this page on the website: http://swift.org/install/linux/tarball

And visiting the Ubuntu 20.04 section. The list of packages doesn't include these packages:

```
unzip libcurl4-openssl-dev libpython3.8 libxml2-dev
```

Note that the equivalent Dockerfile for the official swift lists these packages as required here for 6.1:

https://github.com/swiftlang/swift-docker/blob/main/6.1/ubuntu/20.04/Dockerfile

## Expected behavior

It's expected that the list of required system packages agree between swiftly, the docker files for official swift images, and the website.

## Actual behavior

There is a discrepancy between package lists.

PR #1447, opened 2026-06-17
Fix Fedora 39 package list to match Swift 6.1 Docker image

krishnapermi:
While cross-checking the Linux distro pages against the official Swift Docker images (following up on #954), I noticed the Fedora 39 page has `zip` listed as a required package, but the Swift 6.1 Fedora 39 Docker image only uses `unzip`. I also noticed `libstdc++-devel` and `libstdc++-static` are both in the Docker image but missing from the page. This updates the package list to match.

### DIFF
--- a/_includes/linux/fedora39.html
+++ b/_includes/linux/fedora39.html
@@ -10,6 +10,7 @@ $ yum install \
       libxml2-devel \
       python3-devel \
       sqlite-devel \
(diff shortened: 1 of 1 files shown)


krishnapermi:
Closing in favor of #1459 which covers the same fix with a cleaner description.

PR #1467, opened 2026-06-24
Fix Document class example missing initializer

### LINKED ISSUE #1268: Error in "Value and Reference Types" article: class Document missing initializer

krishnapermi:
The reference-type example uses `Document(text:)` but classes don't get a synthesized memberwise initializer the way structs do - so the code doesn't actually compile. Added an explicit `init(text:)` to fix it.

Closes #1268

### DIFF
--- a/documentation/articles/value-and-reference-types.md
+++ b/documentation/articles/value-and-reference-types.md
@@ -74,6 +74,10 @@ Since it is a reference to the same instance, changing the `text` of `friendDoc`
 ```swift
 class Document {
   var text: String
+
+  init(text: String) {
+    self.text = text
+  }
 }
 
 var myDoc = Document(text: "Great new article")


dempseyatgithub:
Thank you @krishnapermi for your contribution.

This pull request supersedes #1315.

dempseyatgithub (approved):
Tested the updated code example in a playground and it compiles and works as expected.

PR #1484, opened 2026-08-02
Android getting started: just hard-code the last patch release version

finagolfin:
Automating it wasn't working, and this will no longer be needed with the upcoming 6.4 release.


finagolfin:
@heckj, this currently still incorrectly reads `swift sdk remove _android`, ie this previous version code I added and am now removing returns a blank result. I'm not going to dig into whatever weird logic Jekyll's language requires here, which apparently nobody knows, particularly since we'll be dropping this line with the next release.

finagolfin:
@alexandersandberg, simple review

finagolfin:
@daveverwer, could you take a look? Quick review

shahmishal (approved):

PR #1487, opened 2026-08-05
[cxx-interop] Fix a code snippet on non-copyable structs

egorzhdan:
The C++ type that was intended to demonstrate the support for `SWIFT_NONCOPYABLE` did not actually get imported into Swift, because it had implicitly-deleted copy/move constructors due to the rule of five.

rdar://183960967

### DIFF
--- a/documentation/cxx-interop/index.md
+++ b/documentation/cxx-interop/index.md
@@ -331,9 +331,12 @@ makes Swift import it as `~Copyable`, while leaving the type copyable in C++.
 ```c++
 struct SWIFT_NONCOPYABLE FileDescriptor {
   FileDescriptor(const char *path);
-  ~FileDescriptor();
 };
 ```
+
+A type annotated with `SWIFT_NONCOPYABLE` must still be movable in C++. This may
+require explicitly declaring the move constructor in cases where it would
+otherwise be implicitly deleted.
 
 Class templates may need more control over how copyability is determined for each specialization. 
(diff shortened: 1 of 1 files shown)


egorzhdan:
cc @susmonteiro @Xazax-hun @j-hui @patrykstefanski

j-hui:
Would it be clearer if we spelled out the move constructor, since that's the specific 
member that ClangImporter is looking for?

egorzhdan:
@j-hui this example is described as:

> For example, this `FileDescriptor` has an implicit copy constructor ...

If we spell out the move constructor, we'd have to spell out the copy constructor too, due to the rule of five.

j-hui:
Ah got it, I missed that.

Then maybe we should be more explicit in saying:

```
A type annotated with `SWIFT_NONCOPYABLE` must still be movable in C++.
This may require explicitly declaring the move constructor in cases where it would otherwise be implicitly deleted.
```

j-hui (approved):
I don't feel especially strongly about adding that clarification though. This looks good to me as is!

egorzhdan:
I like that suggestion, added the second sequence to the paragraph.

Xazax-hun (approved):
LG!

egorzhdan:
cc @swiftlang/website-workgroup could we please get an approval on this?

heckj (approved):

PR #1490, opened 2026-08-08
Don't specify server code in task stealers mention

ktoso:
Since this feature is darwin platforms only, it is not necessarily clear that it benefits server users -- let's avoid the distinction to not confuse readers.

Great post otherwise, thank you!


davelester (approved):
LGTM, thank you!

PR #1495, opened 2026-08-11
[cxx-interop] Update doc for foreign reference type inheritance

egorzhdan:
Swift infers that derived types are foreign reference types when the base type is annotated with `SWIFT_UNSAFE_REFERENCE`, or `SWIFT_IMMORTAL_REFERENCE`, unlike what the documentation claimed.

rdar://184040817

### DIFF
--- a/documentation/cxx-interop/index.md
+++ b/documentation/cxx-interop/index.md
@@ -1393,6 +1393,15 @@ object2.doSomething()
 // `object` will be released here.
 ```
 
+When a C++ type inherits from a base type that is annotated as a reference type, the Swift compiler automatically infers that the derived type is also a reference type.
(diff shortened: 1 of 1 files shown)


egorzhdan:
cc @j-hui @susmonteiro @Xazax-hun @patrykstefanski

susmonteiro (approved):

Xazax-hun (approved):
LG! I wonder if we should mention that inference only kicks in if there is a unique shared reference base. E.g., multiple different shared reference bases might also not work.

egorzhdan:
> multiple different shared reference bases might also not work

That scenario works in Swift 6.4, the compiler only rejects multiple shared reference bases if they have different retain/release ops.

heckj (approved):

PR #1501, opened 2026-08-18
[Blog post] A blog post about module-related debug info improvements

adrian-prantl:
in Swift 6.4 and a guide for how users who maintain their own build systems can simplify their dependency tracking to take advantage of the improved module tracking.


_posts/2026-08-25-module-tracking-in-debug-info.md
```diff
@@ -0,0 +1,111 @@
...
+# Modulewrap and linker invocation on Linux
+swift-frontend -modulewrap Foo.swiftmodule -o Foo.swiftmodule.o
+lld Foo.o Foo.swiftmodule.o -o MyApplication
+```
+
+This can create scalability issues, especially for large applications:
+- Module files can get large and for an entire application you can often end up with a large portion of the SDK in your Linux or Windows binary. That can be quite problematic for the binary size.
+- As mentioned above, the chances of LLDB finding the right module in a Swift AST section or symbol table just by its base name diminish as the application gets more complex.
+- Binary Swift modules are version locked to the precise compiler that created them. This is at odds with the intent of dSYM bundles, which are meant for long-term archival serialization of debug info.
+- When anything goes wrong and a matching explicit module cannot be found LLDB falls back to an implicit module import which may involve recompiling parts of the SDK from source. The can be very slow.
```
heckj:
```suggestion
- When anything goes wrong and a matching explicit module cannot be found LLDB falls back to an implicit module import which may involve recompiling parts of the SDK from source. This can be very slow.
```

_posts/2026-08-25-module-tracking-in-debug-info.md
```diff
@@ -0,0 +1,111 @@
...
+### New: Swift driver passes module path to compile jobs
+Users of `swiftpm` or Xcode do not need to think about this, because the Swift driver also knows about the new `-debug-module-path` option and automatically passes the path to the own Swift module to the compiler. However, users maintaining their own 3rd party build system to orchestrate Swift compilations with explicitly-built modules need to make sure to communicate the path to the top level module to each object file compilation job.
+### Deprecated: swiftc -modulewrap and ld -add_ast_path
+Because the module paths are now communicated via debug info and the module headers themselves, 3rd build systems doing explicit module builds can now remove all `-modulewrap` actions on Linux and Windows; and remove the use of the `-add_ast_path` linker option on Darwin (macOS, iOS, *etc…*).
```
heckj:
```suggestion
Because the module paths are now communicated via debug info and the module headers themselves, 3rd party build systems doing explicit module builds can now remove all `-modulewrap` actions on Linux and Windows; and remove the use of the `-add_ast_path` linker option on Darwin (macOS, iOS, *etc…*).
```

_posts/2026-08-25-module-tracking-in-debug-info.md
```diff
@@ -0,0 +1,111 @@
...
+### Deprecated: Binary Swift modules dSYM bundles
```
heckj:
```suggestion
### Deprecated: Binary Swift modules in dSYM bundles
```

_posts/2026-08-25-module-tracking-in-debug-info.md
```diff
@@ -0,0 +1,111 @@
...
+
+## New: Precise module tracking
+
+Over the last couple of months we have been making changes to the Swift compiler, the Swift driver, and LLDB that improve performance, reliability, and scalability. These changes were enabled by and are built on top of explicitly-built modules.
+### New: Explicitly-built modules track their explicit Clang and Swift dependencies
+Explicitly-built binary Swift modules have always kept track of their explicitly-built Clang module dependencies. This is what makes importing explicit modules in LLDB so fast. In Swift 6.3, explicitly-built binary Swift modules also store their *Swift* module dependencies. This makes importing an explicitly-built module fast and unambiguous because no module needs to be looked up by name. This just happens automatically, users don’t need to make any changes. Users with distributed build systems will already be familiar with the Swift frontend’s path remapping options, which now also affect Swift module paths.
+### New: Debug info stores path of own Swift module
```
heckj:
```suggestion
### New: Debug info stores path of its own Swift module
```

_posts/2026-08-25-module-tracking-in-debug-info.md
```diff
@@ -0,0 +1,111 @@
...
+Once LLDB found the top-level module it can precisely import it and all of its dependencies. But how can LLDB find precisely the module that belongs to the Swift file at the current breakpoint? In Swift 6.3, the Swift compiler can store that path to it in the debug info. Because a Swift file’s own Swift module is not an input to an object file compilation, there is a new `-debug-module-path` compiler option to communicate the path to each object file compilation action. This path is also subject to the standard path remapping options used by users with distributed build systems.
+### New: Swift driver passes module path to compile jobs
+Users of `swiftpm` or Xcode do not need to think about this, because the Swift driver also knows about the new `-debug-module-path` option and automatically passes the path to the own Swift module to the compiler. However, users maintaining their own 3rd party build system to orchestrate Swift compilations with explicitly-built modules need to make sure to communicate the path to the top level module to each object file compilation job.
```
heckj:
```suggestion
Users of `swiftpm` or Xcode do not need to think about this, because the Swift driver also knows about the new `-debug-module-path` option and automatically passes the path of its own Swift module to the compiler. However, users maintaining their own 3rd party build system to orchestrate Swift compilations with explicitly-built modules need to make sure to communicate the path to the top-level module to each object file compilation job.
```

fix for missing/incorrect grammar "of it's own"
top level -> top-level for consistency with usage earlier

_posts/2026-08-25-module-tracking-in-debug-info.md
```diff
@@ -0,0 +1,111 @@
...
+- When anything goes wrong and a matching explicit module cannot be found LLDB falls back to an implicit module import which may involve recompiling parts of the SDK from source. The can be very slow.
+
+## New: Precise module tracking
+
+Over the last couple of months we have been making changes to the Swift compiler, the Swift driver, and LLDB that improve performance, reliability, and scalability. These changes were enabled by and are built on top of explicitly-built modules.
+### New: Explicitly-built modules track their explicit Clang and Swift dependencies
+Explicitly-built binary Swift modules have always kept track of their explicitly-built Clang module dependencies. This is what makes importing explicit modules in LLDB so fast. In Swift 6.3, explicitly-built binary Swift modules also store their *Swift* module dependencies. This makes importing an explicitly-built module fast and unambiguous because no module needs to be looked up by name. This just happens automatically, users don’t need to make any changes. Users with distributed build systems will already be familiar with the Swift frontend’s path remapping options, which now also affect Swift module paths.
```
heckj:
```suggestion
Explicitly-built binary Swift modules have always kept track of their explicitly-built Clang module dependencies. This is what makes importing explicit modules in LLDB so fast. In Swift 6.3, explicitly-built binary Swift modules also store their *Swift* module dependencies. This makes importing an explicitly-built module fast and unambiguous because no module needs to be looked up by name. This just happens automatically. Users don’t need to make any changes. Users with distributed build systems will already be familiar with the Swift frontend’s path remapping options, which now also affect Swift module paths.
```

_posts/2026-08-25-module-tracking-in-debug-info.md
```diff
@@ -0,0 +1,111 @@
...
+If the absence of binary Swift modules in dSYM bundles creates an unforeseen problem with your workflow, please let us know, either on the [Swift LLDB forum](https://forums.swift.org/c/development/lldb/13) or by creating an issue on the [bug tracker](https://github.com/swiftlang/swift/issues)!
+### What about caching?
+All the paths stored in Swift modules and debug info can also be CAS addresses, so all of the described features also transparently work with compilation caching.
+### Coming in Swift 6.4: Faster bridging header import in LLDB
+Up to including Swift 6.3, LLDB always compiles a bridging header from source. In recent nightly development toolchains, LLDB can use the new precise explicit module information to import precompiled bridging headers and their explicit module dependencies directly. This makes debugging explicitly built projects with bridging headers as fast and reliable as debugging fully modularized projects.
```
heckj:
```suggestion
Up to including Swift 6.3, LLDB always compiles a bridging header from source. In recent nightly development toolchains, LLDB can use the new precise explicit module information to import precompiled bridging headers and their explicit module dependencies directly. This makes debugging explicitly-built projects with bridging headers as fast and reliable as debugging fully modularized projects.
```

other reference in the article use hyphenated "explicitly-built", so suggesting we apply it here too

_posts/2026-08-25-module-tracking-in-debug-info.md
```diff
@@ -0,0 +1,111 @@
...
+```bash
+# Modulewrap and linker invocation on Linux
+swift-frontend -modulewrap Foo.swiftmodule -o Foo.swiftmodule.o
+lld Foo.o Foo.swiftmodule.o -o MyApplication
+```
+
+This can create scalability issues, especially for large applications:
+- Module files can get large and for an entire application you can often end up with a large portion of the SDK in your Linux or Windows binary. That can be quite problematic for the binary size.
+- As mentioned above, the chances of LLDB finding the right module in a Swift AST section or symbol table just by its base name diminish as the application gets more complex.
+- Binary Swift modules are version locked to the precise compiler that created them. This is at odds with the intent of dSYM bundles, which are meant for long-term archival serialization of debug info.
```
heckj:
```suggestion
- Binary Swift modules are version-locked to the precise compiler that created them. This is at odds with the intent of dSYM bundles, which are meant for long-term archival serialization of debug info.
```

I think version-locked is correct (hyphenated)

_posts/2026-08-25-module-tracking-in-debug-info.md
```diff
@@ -0,0 +1,111 @@
...
+## Swift modules, debug info, and the build system
+
+Let’s look at where those modules are found next. In order to communicate the location of `Foo.swiftmodule` to LLDB, Swift build systems rely on some cooperation from the linker. On Darwin the system linker accepts an option called `-add_ast_path` and build systems are expected to use this option to list every binary Swift module when linking.
+
+```bash
+# Linker invocation on macOS
+ld -add_ast_path /path/to/Foo.swiftmodule Foo.o -o MyApplication
+```
+
+The linker translates these options into symbol table entries. The debug info linker dsymutil then collects all Swift modules and stores them in a special `__swift_ast` section in the dSYM bundle, where LLDB can find them by name. Alternatively, when debugging without dSYM bundles, LLDB reads the symbol table entries in the binary to collect a list of all binary Swift modules.
```
heckj:
```suggestion
The linker translates these options into symbol table entries. The debug info linker `dsymutil` then collects all Swift modules and stores them in a special `__swift_ast` section in the dSYM bundle, where LLDB can find them by name. Alternatively, when debugging without dSYM bundles, LLDB reads the symbol table entries in the binary to collect a list of all binary Swift modules.
```

marking dsymutil with backticks to put it consistently in code voice to match it's other usage in this article

_posts/2026-08-25-module-tracking-in-debug-info.md
```diff
@@ -0,0 +1,111 @@
...
+In order to JIT-compile user expressions that make use of data types defined in the debugged program, LLDB’s embedded Swift compiler needs to import the Swift modules defining those types. In a world before explicitly-built modules, LLDB would find the base name of the main module at the current breakpoint in the debug info and then kick off an implicit import of a module with that name. With a cold module cache this would launch an expensive compilation of that module and all its dependencies.
+
+To illustrate this, let’s walk through a simple example:
+
+```swift
+(lldb) p myObj
+```
+
+Here myObj is just a local variable: LLDB can find its location in the debug info and resolve its type via reflection metadata. No need to bother the Swift compiler.
```
heckj:
```suggestion
Here `myObj` is just a local variable: LLDB can find its location in the debug info and resolve its type via reflection metadata. No need to bother the Swift compiler.
```

marking `myObj` as "code voice" since it's a symbol reference

_posts/2026-08-25-module-tracking-in-debug-info.md
```diff
@@ -0,0 +1,111 @@
...
+### New: Debug info stores path of own Swift module
+Once LLDB found the top-level module it can precisely import it and all of its dependencies. But how can LLDB find precisely the module that belongs to the Swift file at the current breakpoint? In Swift 6.3, the Swift compiler can store that path to it in the debug info. Because a Swift file’s own Swift module is not an input to an object file compilation, there is a new `-debug-module-path` compiler option to communicate the path to each object file compilation action. This path is also subject to the standard path remapping options used by users with distributed build systems.
```
heckj:
```suggestion
Once LLDB found the top-level module it can precisely import it and all of its dependencies. But how can LLDB find precisely the module that belongs to the Swift file at the current breakpoint? In Swift 6.3, the Swift compiler can store the path in the debug info. Because a Swift file’s own Swift module is not an input to an object file compilation, there is a new `-debug-module-path` compiler option to communicate the path to each object file compilation action. This path is also subject to the standard path remapping options used by users with distributed build systems.
```

optional rewording suggestion - I think this reads slightly simpler, as long as it's still correct.

heckj (changes requested):
A few grammar and spelling nits and mistakes with suggestions for fixing them, as well as a single optional reword to make a sentence a bit easier to read and parse (which I *think* is still technically correct, but your call if it makes more sense the way it is)

_posts/2026-08-25-module-tracking-in-debug-info.md
davelester: @heckj @adrian-prantl I'd suggest writing out "third-party" (hyphenated)

_posts/2026-08-25-module-tracking-in-debug-info.md
adrian-prantl: Here "own Swift module" is (weak) a term of art. I replaced it with "object file's own Swift module". Does that read better?

_posts/2026-08-25-module-tracking-in-debug-info.md
adrian-prantl: same edit as above

_posts/2026-08-25-module-tracking-in-debug-info.md
```diff
@@ -0,0 +1,111 @@
...
+This article explains how Swift modules are used by the debugger, and how they are connected with debug info. Next, it explains how Swift 6.3+ changes how modules are tracked in debug info to solve several problems with the previous representation. Finally, it shows how to adjust build systems to make use of the new representation, and eliminate some build steps that are no longer necessary.
+
+## Swift modules and expression evaluation
+
+LLDB’s standout feature is its [powerful expression evaluator](https://lldb.llvm.org/#compiler-integration-benefits). Because LLDB embeds the Clang and Swift compilers, it can JIT-compile any valid source code and run it in the context of your application while stopped at a breakpoint. This includes not just calling code in your application, but also defining new data types, functions, and closures. Debugging features that are usually reserved for interpreted or JIT-compiled languages like Python become available to ahead-of-time-compiled languages like C++ and, of course, Swift!
```
augusto2112: https://lldb.llvm.org/#compiler-integration-benefits only talks about clang and clang types. Not sure if the link is needed here as it doesn't talk about expression evaluation in general terms, only in the context of clang.

augusto2112: "it can JIT-compile any valid source code and run..." I'm not sure if JIT compilation is something most developers know about. I think just compile reads fine in this sentence. Or introduce the term "JIT-compilation"

augusto2112: "Debugging features that are usually reserved for interpreted or JIT-compiled languages like Python" first time I read this I thought you meant that Python was JIT-compiled.  Re-reading I understand that you meant that Python is an example of an interpreted language, but I'd either replace the language example with something like Javascript where every implementation is JIT compiled.

_posts/2026-08-25-module-tracking-in-debug-info.md
```diff
@@ -0,0 +1,111 @@
...
+- When anything goes wrong and a matching explicit module cannot be found LLDB falls back to an implicit module import which may involve recompiling parts of the SDK from source. This can be very slow.
+
+## New: Precise module tracking
+
+Over the last couple of months we have been making changes to the Swift compiler, the Swift driver, and LLDB that improve performance, reliability, and scalability. These changes were enabled by and are built on top of explicitly-built modules.
+### New: Explicitly-built modules track their explicit Clang and Swift dependencies
+Explicitly-built binary Swift modules have always kept track of their explicitly-built Clang module dependencies. This is what makes importing explicit modules in LLDB so fast. In Swift 6.3, explicitly-built binary Swift modules also store their *Swift* module dependencies. This makes importing an explicitly-built module fast and unambiguous because no module needs to be looked up by name. This just happens automatically. Users don’t need to make any changes. Users with distributed build systems will already be familiar with the Swift frontend’s path remapping options, which now also affect Swift module paths.
```
augusto2112: I think "This is what makes importing explicit modules in LLDB so fast." should be removed or moved to after talking about the new part. Also, I think you should emphasize that before there was no explicit swift module tracking. Something like:

Explicitly-built binary Swift modules have always kept track of their explicitly-built Clang module dependencies. **However, explicit Swift modules were not tracked.** ...

Also, the title should be "Explicitly-built modules track their explicit  Swift dependencies" since only Swift modules are new.

_posts/2026-08-25-module-tracking-in-debug-info.md
```diff
@@ -0,0 +1,111 @@
...
+#### But don’t we need them for debugging?
+Since Swift 1.0, binary Swift modules were included in dSYM bundles because they were needed to resolve the types of local variables. However, starting with Swift 5.6, LLDB could perform this operation by reading the reflection metadata in the binary. The absence of binary Swift modules in dSYM bundles does not affect LLDB’s ability to inspect the contents of variables. Binary Swift modules are still needed to evaluate complex expressions. Expression evaluation continues to work as long as LLDB finds all binary modules in their original (or [remapped](https://lldb.llvm.org/use/map.html#remap-source-file-pathnames-for-the-debug-session)) location. This is always the case when debugging a just-built binary on the same machine.
```
augusto2112: What is "evaluate complex expressions"? Spell out that these are function calls, computed getters, modifying program state?

_posts/2026-08-25-module-tracking-in-debug-info.md
```diff
@@ -0,0 +1,111 @@
...
+If the absence of binary Swift modules in dSYM bundles creates an unforeseen problem with your workflow, please let us know, either on the [Swift LLDB forum](https://forums.swift.org/c/development/lldb/13) or by creating an issue on the [bug tracker](https://github.com/swiftlang/swift/issues)!
+### What about caching?
+All the paths stored in Swift modules and debug info can also be CAS addresses, so all of the described features also transparently work with compilation caching.
```
augusto2112, line 98: We never talked about what CAS is. I think at least a link would be useful.

_posts/2026-08-25-module-tracking-in-debug-info.md
```diff
@@ -0,0 +1,111 @@
+---
+layout: new-layouts/post
+published: false
+date: 2026-08-25 11:30
+title: "Module Tracking in Swift Debug Info"
+author: [adrianprantl]
+category: "Developer Tools"
+---
+Starting with Swift 6.3, and completed in 6.4, there are some changes coming to how the Swift compiler references explicitly-built Swift modules in debug info. The majority of developers will automatically benefit from **faster**, more **reliable debugging** and **smaller build products**, without any modifications to their SwiftPM or Xcode projects. For developers who **maintain their own build systems** using, for example, Bazel, Buck, or CMake, **some adjustments may be necessary** to take advantage of these changes.
+This article explains how Swift modules are used by the debugger, and how they are connected with debug info. Next, it explains how Swift 6.3+ changes how modules are tracked in debug info to solve several problems with the previous representation. Finally, it shows how to adjust build systems to make use of the new representation, and eliminate some build steps that are no longer necessary.
```
JDevlieghere:
```suggestion
This article explains how Swift modules are used by the debugger, and how they relate to debug info. Next, it explains how Swift 6.3+ changes how modules are tracked in debug info to solve several problems with the previous representation. Finally, it shows how to adjust build systems to make use of the new representation, and eliminate some build steps that are no longer necessary.
```

JDevlieghere: "are connected with" sounds slightly odd. What about "relate to" or "are represented in"?

_posts/2026-08-25-module-tracking-in-debug-info.md
```diff
@@ -0,0 +1,111 @@
+---
+layout: new-layouts/post
+published: false
+date: 2026-08-25 11:30
+title: "Module Tracking in Swift Debug Info"
+author: [adrianprantl]
+category: "Developer Tools"
+---
+Starting with Swift 6.3, and completed in 6.4, there are some changes coming to how the Swift compiler references explicitly-built Swift modules in debug info. The majority of developers will automatically benefit from **faster**, more **reliable debugging** and **smaller build products**, without any modifications to their SwiftPM or Xcode projects. For developers who **maintain their own build systems** using, for example, Bazel, Buck, or CMake, **some adjustments may be necessary** to take advantage of these changes.
```
JDevlieghere, line 9: This whole article assumes the reader is familiar with explicitly-built Swift modules. Is there a blogpost we can link?

_posts/2026-08-25-module-tracking-in-debug-info.md
```diff
@@ -0,0 +1,111 @@
...
+```
+
+Here `myObj` is just a local variable: LLDB can find its location in the debug info and resolve its type via reflection metadata. No need to bother the Swift compiler.
+Let’s make it more complex:
+
+```swift
+(lldb) p myObj.myComputedProperty
+```
+
+Here, `myComputedProperty` is a function call; in order to evaluate this, LLDB needs the expression evaluator to run code in the target. In order to initialize a Swift compiler instance with the state of the current module, LLDB finds the name of the current function’s Swift module in debug info.
```
JDevlieghere: Nit: the previous paragraph also started with "Here". Maybe something like "In this case"?

JDevlieghere: Suggestion: is a function call -> "is really a function call"?

_posts/2026-08-25-module-tracking-in-debug-info.md
```diff
@@ -0,0 +1,111 @@
...
+  // Expression begins here:
+  myObj.myComputedProperty
+  ...
+```
+
+One problem with this is that `import Foo` is quite imprecise: Even though the Swift language doesn’t allow multiple modules to have the same name, even the most stringently engineered application may have more than one copy of the same module. For example, there might be a private version of a module containing all of its private declarations (which would be great for LLDB) and also a Swift interface file that only contains the public interface for the module. Or there might be `macOS` and `ios-macabi` variants of the same module in the same process.
+
+## Swift modules, debug info, and the build system
+
+Let’s look at where those modules are found next. In order to communicate the location of `Foo.swiftmodule` to LLDB, Swift build systems rely on some cooperation from the linker. On Darwin the system linker accepts an option called `-add_ast_path` and build systems are expected to use this option to list every binary Swift module when linking.
```
JDevlieghere:
```suggestion
Let’s look at where those modules are found next. In order to communicate the location of `Foo.swiftmodule` to LLDB, Swift build systems rely on some cooperation from the linker. On Darwin, the system linker accepts an option called `-add_ast_path` and build systems are expected to specify this option to list every binary Swift module when linking.
```

JDevlieghere (commented):
I didn't get to the end yet, but submitting the comments I have so far.

_posts/2026-08-25-module-tracking-in-debug-info.md
heckj: perfect, thank you

heckj (approved):

_posts/2026-08-25-module-tracking-in-debug-info.md
adrian-prantl: Replaced Python with JavaScript.

_posts/2026-08-25-module-tracking-in-debug-info.md
adrian-prantl: Clarified.

_posts/2026-08-25-module-tracking-in-debug-info.md
adrian-prantl: I have not found a blog post. There is this older forum thread: https://forums.swift.org/t/evolving-how-swift-is-built-driver-as-a-library-explicitly-built-modules-and-beyond/61102 but that is more aimed at compiler developers.

_posts/2026-08-25-module-tracking-in-debug-info.md
adrian-prantl: Similar problem as above, we never blogged about compiler caching either...

_posts/2026-08-25-module-tracking-in-debug-info.md
```diff
@@ -0,0 +1,111 @@
...
+import Foo
+
+func lldb_expr(_ $__lldb_arg : UnsafeMutablePointer<Any>) {
+  let myObj: MyObject = /* some LLDB magic */
+  // Expression begins here:
+  myObj.myComputedProperty
+  ...
+```
+
+One problem with this is that `import Foo` is quite imprecise: Even though the Swift language doesn’t allow multiple modules to have the same name, even the most stringently engineered application may have more than one copy of the same module. For example, there might be a private version of a module containing all of its private declarations (which would be great for LLDB) and also a Swift interface file that only contains the public interface for the module. Or there might be `macOS` and `ios-macabi` variants of the same module in the same process.
```
owenv, line 55: `macOS` and `Mac Catalyst` might be a little more familiar terminology for most users

_posts/2026-08-25-module-tracking-in-debug-info.md
```diff
@@ -0,0 +1,111 @@
...
+### New: Debug info stores path of object file's own Swift module
+Once LLDB found the top-level module it can precisely import it and all of its dependencies. But how can LLDB find precisely the module that belongs to the Swift file at the current breakpoint? In Swift 6.3, the Swift compiler can store the path to it in the debug info. Because a Swift file’s own Swift module is not an input to an object file compilation, there is a new `-debug-module-path` compiler option to communicate the path to each object file compilation action. This path is also subject to the standard path remapping options used by users with distributed build systems.
```
owenv, line 87: Once LLDB found -> "Once LLDB has found" or "Once LLDB finds"

owenv (approved):

_posts/2026-08-25-module-tracking-in-debug-info.md
```diff
@@ -0,0 +1,111 @@
...
+
+## New: Precise module tracking
+
+Over the last couple of months we have been making changes to the Swift compiler, the Swift driver, and LLDB that improve performance, reliability, and scalability. These changes were enabled by and are built on top of explicitly-built modules.
+### New: Explicitly-built modules track their explicit Swift dependencies
+Explicitly-built binary Swift modules have always kept track of their explicitly-built *Clang* module dependencies. This is why LLDB can import explicit modules so much faster than implicit modules, which may need to recompile their dependencies from source. In Swift 6.3, explicitly-built binary Swift modules also store their *Swift* module dependencies. This makes importing an explicitly-built module fast and unambiguous because no module needs to be looked up by name. This just happens automatically. Users don’t need to make any changes. Users with distributed build systems will already be familiar with the Swift frontend’s path remapping options, which now also affect Swift module paths.
```
artemcm: > binary Swift modules also store their *Swift* module dependencies

This should be a bit more precise to point out that Swift modules store info or location of their *Swift* module dependencies, or "keep track of" their *Swift* dependencies, like in the sentence above.

artemcm:
```suggestion
Explicitly-built binary Swift modules have always kept track of their explicitly-built *Clang* module dependencies. This is why LLDB can import explicit modules so much faster than implicit modules, which may need to recompile their dependencies from source. In Swift 6.3, explicitly-built binary Swift modules also store their *Swift* module dependencies. This makes importing an explicitly-built module fast and unambiguous because no module needs to be looked up by name. This happens automatically. Users don’t need to make any changes. Users with distributed build systems will already be familiar with the Swift frontend’s path remapping options, which now also affect Swift module paths.
```

_posts/2026-08-25-module-tracking-in-debug-info.md
```diff
@@ -0,0 +1,111 @@
...
+### New: Debug info stores path of object file's own Swift module
+Once LLDB found the top-level module it can precisely import it and all of its dependencies. But how can LLDB find precisely the module that belongs to the Swift file at the current breakpoint? In Swift 6.3, the Swift compiler can store the path to it in the debug info. Because a Swift file’s own Swift module is not an input to an object file compilation, there is a new `-debug-module-path` compiler option to communicate the path to each object file compilation action. This path is also subject to the standard path remapping options used by users with distributed build systems.
+### New: Swift driver passes module path to compile jobs
+Users of `swiftpm` or Xcode do not need to think about this, because the Swift driver also knows about the new `-debug-module-path` option and automatically passes the path to the object file's own Swift module to the compiler. However, users maintaining their own third-party build system to orchestrate Swift compilations with explicitly-built modules need to make sure to communicate the path to the top-level module to each object file compilation job.
```
artemcm: How would they do this? Why can't the driver do this automatically for them?

adrian-prantl: clarified.

_posts/2026-08-25-module-tracking-in-debug-info.md
```diff
@@ -0,0 +1,111 @@
...
+## Summary
+
+With these changes for **explicitly-built modules**:
+- Binaries built with debug info on Windows and Linux, and dSYM bundles on Darwin will get dramatically smaller, since they no longer contain any binary Swift modules (6.4+)
+- Contextual module imports in LLDB become more reliable due to precise tracking instead of by-name lookups
+- Certain performance cliffs around module importing in LLDB are eliminated (such as SDK module dependencies in dSYMs triggering implicit imports)
+- **Developers maintaining their own build systems** can remove support for `-modulewrap` actions and remove `-add_ast_path` from the linker flags, but may need to pass `-debug-module-path` to the compiler if they are not letting the Swift driver handle the frontend options
+- Finally, many projects forgot to use `-add_ast_path` when linking static archives into their projects and had mysterious debugging problems inside the static archives as a result. This entire class of problems has been designed away.
+
+*tl;dr:* `-module-wrap` and `-add_ast_path` are replaced by `-debug-module-path`. Debug info gets smaller and more precise!
```
davelester:
```suggestion
*tl;dr:* `-modulewrap` and `-add_ast_path` are replaced by `-debug-module-path`. Debug info gets smaller and more precise!
```

Should this be `-modulewrap` to be consistent with previous mentions in the post?

_posts/2026-09-11-module-tracking-in-debug-info.md
```diff
@@ -0,0 +1,111 @@
...
+```bash
+# Linker invocation on macOS
+ld -add_ast_path /path/to/Foo.swiftmodule Foo.o -o MyApplication
+```
+
+The linker translates these options into symbol table entries. The debug info linker `dsymutil` then collects all Swift modules and stores them in a special `__swift_ast` section in the dSYM bundle, where LLDB can find them by name. Alternatively, when debugging without dSYM bundles, LLDB reads the symbol table entries in the binary to collect a list of all binary Swift modules.
+Such an approach would not work on platforms where the linker isn’t aware of Swift. For these platforms the Swift compiler provides a `-modulewrap` action that takes a binary Swift module and outputs an object file with a `.swift_ast` section holding the contents of the module. This object file can then be passed to any linker to get added to the binary, where LLDB can find it.
+
+```bash
+# Modulewrap and linker invocation on Linux
```
davelester, line 77: Could the same apply to Windows? Maybe a quick note next to this code example would cover it.

_posts/2026-08-25-module-tracking-in-debug-info.md
```diff
@@ -0,0 +1,111 @@
...
+
+This can create scalability issues, especially for large applications:
+- Module files can get large and for an entire application you can often end up with a large portion of the SDK in your Linux or Windows binary. That can be quite problematic for the binary size.
+- As mentioned above, the chances of LLDB finding the right module in a Swift AST section or symbol table just by its base name diminish as the application gets more complex.
+- Binary Swift modules are version-locked to the precise compiler that created them. This is at odds with the intent of dSYM bundles, which are meant for long-term archival serialization of debug info.
+- When anything goes wrong and a matching explicit module cannot be found LLDB falls back to an implicit module import which may involve recompiling parts of the SDK from source. This can be very slow.
+
+## New: Precise module tracking
+
+Over the last couple of months we have been making changes to the Swift compiler, the Swift driver, and LLDB that improve performance, reliability, and scalability. These changes were enabled by and are built on top of explicitly-built modules.
```
davelester: It'd be great to transition into this section with a quick statement of a problem that's being solved, in addition to the existing language of performance, reliability, and scalability. Would that maintain its meaning, while making it a bit more approachable?

Suggestion below:

```suggestion
When Swift builds a module, the debugger later needs to know exactly where that module's data lives, and until now that information was tracked loosely enough that distributed or path-remapping build systems could lose track of the exact module. Over the last couple of months we have been making changes to the Swift compiler, the Swift driver, and LLDB that address this directly, while improving performance, reliability, and scalability. These changes were enabled by and are built on top of explicitly-built modules.
```

_posts/2026-08-25-module-tracking-in-debug-info.md
```diff
@@ -0,0 +1,111 @@
...
+If the absence of binary Swift modules in dSYM bundles creates an unforeseen problem with your workflow, please let us know, either on the [Swift LLDB forum](https://forums.swift.org/c/development/lldb/13) or by creating an issue on the [bug tracker](https://github.com/swiftlang/swift/issues)!
+### What about caching?
+All the paths stored in Swift modules and debug info can also be CAS addresses, so all of the described features also transparently work with compilation caching.
+### Coming in Swift 6.4: Faster bridging header import in LLDB
+Up to including Swift 6.3, LLDB always compiles a bridging header from source. In recent nightly development toolchains, LLDB can use the new precise explicit module information to import precompiled bridging headers and their explicit module dependencies directly. This makes debugging explicitly-built projects with bridging headers as fast and reliable as debugging fully modularized projects.
```
davelester: Added a quick transition sentence to make this more approachable.

```suggestion
Beyond more reliable path tracking, 6.4 will also speed up importing bridging headers, a step common enough across Swift projects that most developers will feel the difference.

Up to and including Swift 6.3, LLDB always compiles a bridging header from source, a step that can add noticeable time to debugging sessions that use one. In recent nightly development toolchains, LLDB can use the new precise explicit module information to import precompiled bridging headers and their explicit module dependencies directly. This makes debugging explicitly-built projects with bridging headers as fast and reliable as debugging fully modularized projects.
```

_posts/2026-08-25-module-tracking-in-debug-info.md
```diff
@@ -0,0 +1,111 @@
+---
+layout: new-layouts/post
+published: false
+date: 2026-08-25 11:30
```
davelester:
```suggestion
date: 2026-09-08 13:30
```

Possible revision to Tuesday, if further revisions are needed then we'll update to Thursday.

_posts/2026-08-25-module-tracking-in-debug-info.md
```diff
@@ -0,0 +1,111 @@
...
+lld Foo.o Foo.swiftmodule.o -o MyApplication
+```
+
+This can create scalability issues, especially for large applications:
+- Module files can get large and for an entire application you can often end up with a large portion of the SDK in your Linux or Windows binary. That can be quite problematic for the binary size.
+- As mentioned above, the chances of LLDB finding the right module in a Swift AST section or symbol table just by its base name diminish as the application gets more complex.
+- Binary Swift modules are version-locked to the precise compiler that created them. This is at odds with the intent of dSYM bundles, which are meant for long-term archival serialization of debug info.
+- When anything goes wrong and a matching explicit module cannot be found LLDB falls back to an implicit module import which may involve recompiling parts of the SDK from source. This can be very slow.
+
+## New: Precise module tracking
```
davelester:
```suggestion
## Precise module tracking
```

_posts/2026-08-25-module-tracking-in-debug-info.md
davelester: Small revisions to opening text + a lead-in sentence before jumping into the topic.

```suggestion
When your Swift program hits a breakpoint and stops so you can inspect it, the debugger has to find the exact Swift module your code was built from. Until now, that lookup wasn't always precise. The upcoming Swift 6.4 release will include changes, begun in Swift 6.3, that address this by updating how the Swift compiler references explicitly-built Swift modules in debug info.

The majority of developers will automatically benefit from **faster**, more **reliable debugging** and **smaller build products**, without any modifications to their SwiftPM or Xcode projects.

For developers who **maintain their own build systems** using, for example, Bazel, Buck, or CMake, **some adjustments may be necessary** to take advantage of these changes.
```

_posts/2026-09-11-module-tracking-in-debug-info.md
```diff
@@ -0,0 +1,111 @@
...
+
+## Swift modules and expression evaluation
+
+LLDB’s standout feature is its [powerful expression evaluator](https://lldb.llvm.org/#compiler-integration-benefits). Because LLDB embeds the Clang and Swift compilers, it can JIT-compile any valid source code and run it in the context of your application while stopped at a breakpoint. This includes not just calling code in your application, but also defining new data types, functions, and closures. Debugging features that are usually reserved for interpreted or JIT-compiled languages like JavaScript become available to ahead-of-time-compiled languages like C++ and, of course, Swift!
+In order to JIT-compile user expressions that make use of data types defined in the debugged program, LLDB’s embedded Swift compiler needs to import the Swift modules defining those types. In a world before explicitly-built modules, LLDB would find the base name of the main module at the current breakpoint in the debug info and then kick off an implicit import of a module with that name. With a cold module cache this would launch an expensive compilation of that module and all its dependencies.
```
davelester, line 22:
```suggestion

In order to JIT-compile user expressions that make use of data types defined in the debugged program, LLDB’s embedded Swift compiler needs to import the Swift modules defining those types. In a world before explicitly-built modules, LLDB would find the base name of the main module at the current breakpoint in the debug info and then kick off an implicit import of a module with that name. With a cold module cache this would launch an expensive compilation of that module and all its dependencies.
```

Adding a newline in the markdown file.

_posts/2026-08-25-module-tracking-in-debug-info.md
davelester: Let's update the heading to remove "New: ", and I'll offer additional suggestions for how we can organize the contents below as  list items to make this easier to browse and more aligned with the styling of other posts on the Swift blog.

_posts/2026-08-25-module-tracking-in-debug-info.md
```diff
@@ -0,0 +1,111 @@
...
+
+Over the last couple of months we have been making changes to the Swift compiler, the Swift driver, and LLDB that improve performance, reliability, and scalability. These changes were enabled by and are built on top of explicitly-built modules.
+### New: Explicitly-built modules track their explicit Swift dependencies
```
davelester:
```suggestion
### What's new

```

davelester: Reorganizing this section + making a dedicated "What's new" and "What's deprecated" section and moving existing title styles as bolded list items.

_posts/2026-08-25-module-tracking-in-debug-info.md
```diff
@@ -0,0 +1,111 @@
...
+Explicitly-built binary Swift modules have always kept track of their explicitly-built *Clang* module dependencies. This is why LLDB can import explicit modules so much faster than implicit modules, which may need to recompile their dependencies from source. In Swift 6.3, explicitly-built binary Swift modules also keep track of their *Swift* module dependencies. This makes importing an explicitly-built module fast and unambiguous because no module needs to be looked up by name. This happens automatically. Users don’t need to make any changes. Users with distributed build systems will already be familiar with the Swift frontend’s path remapping options, which now also affect Swift module paths.
```
davelester:
```suggestion
* __Explicitly-built modules track their explicit Swift dependencies:__ Explicitly-built binary Swift modules have always kept track of their explicitly-built *Clang* module dependencies. This is why LLDB can import explicit modules so much faster than implicit modules, which may need to recompile their dependencies from source. In Swift 6.3, explicitly-built binary Swift modules also keep track of their *Swift* module dependencies. This makes importing an explicitly-built module fast and unambiguous because no module needs to be looked up by name. This happens automatically. Users don’t need to make any changes. Users with distributed build systems will already be familiar with the Swift frontend’s path remapping options, which now also affect Swift module paths.
```

_posts/2026-08-25-module-tracking-in-debug-info.md
```diff
@@ -0,0 +1,111 @@
...
+### New: Debug info stores path of object file's own Swift module
```
davelester:
```suggestion
```

Removing in order to move as bolded title in list item.

_posts/2026-08-25-module-tracking-in-debug-info.md
```diff
@@ -0,0 +1,111 @@
...
+Once LLDB finds the top-level module it can precisely import it and all of its dependencies. But how can LLDB find precisely the module that belongs to the Swift file at the current breakpoint? In Swift 6.3, the Swift compiler can store the path to it in the debug info. Because a Swift file’s own Swift module is not an input to an object file compilation, there is a new `-debug-module-path` compiler option to communicate the path to each object file compilation action. This path is also subject to the standard path remapping options used by users with distributed build systems.
```
davelester:
```suggestion
* __Debug info stores path of object file's own Swift module:__ Once LLDB finds the top-level module it can precisely import it and all of its dependencies. But how can LLDB find precisely the module that belongs to the Swift file at the current breakpoint? In Swift 6.3, the Swift compiler can store the path to it in the debug info. Because a Swift file’s own Swift module is not an input to an object file compilation, there is a new `-debug-module-path` compiler option to communicate the path to each object file compilation action. This path is also subject to the standard path remapping options used by users with distributed build systems.
```

_posts/2026-08-25-module-tracking-in-debug-info.md
```diff
@@ -0,0 +1,111 @@
...
+### New: Swift driver passes module path to compile jobs
```
davelester:
```suggestion
```

Removing in order to become bolded title in list item.

_posts/2026-08-25-module-tracking-in-debug-info.md
```diff
@@ -0,0 +1,111 @@
...
+Users of `swiftpm` or Xcode do not need to think about this, because the Swift driver also knows about the new `-debug-module-path` option and automatically passes the path to the object file's own Swift module to the compiler. However, users maintaining their own third-party build system to orchestrate Swift compilations with explicitly-built modules that are calling the Swift frontend directly and bypassing the Swift driver need to make sure to communicate the path to the top-level module to each object file compilation job.
```
davelester:
```suggestion
* __Swift driver passes module path to compile jobs:__ Users of `swiftpm` or Xcode do not need to think about this, because the Swift driver also knows about the new `-debug-module-path` option and automatically passes the path to the object file's own Swift module to the compiler. However, users maintaining their own third-party build system to orchestrate Swift compilations with explicitly-built modules that are calling the Swift frontend directly and bypassing the Swift driver need to make sure to communicate the path to the top-level module to each object file compilation job.
```

_posts/2026-08-25-module-tracking-in-debug-info.md
```diff
@@ -0,0 +1,111 @@
...
+### Deprecated: swiftc -modulewrap and ld -add_ast_path
```
davelester:
```suggestion

### What's deprecated

```

Reorganization, similar to "What's new" section above.

_posts/2026-08-25-module-tracking-in-debug-info.md
```diff
@@ -0,0 +1,111 @@
...
+Because the module paths are now communicated via debug info and the module headers themselves, third-party build systems doing explicit module builds can now remove all `-modulewrap` actions on Linux and Windows; and remove the use of the `-add_ast_path` linker option on Darwin (macOS, iOS, *etc…*).
```
davelester:
```suggestion
* __swiftc -modulewrap and ld -add_ast_path:__ Because the module paths are now communicated via debug info and the module headers themselves, third-party build systems doing explicit module builds can now remove all `-modulewrap` actions on Linux and Windows; and remove the use of the `-add_ast_path` linker option on Darwin (macOS, iOS, *etc…*).
```

_posts/2026-08-25-module-tracking-in-debug-info.md
```diff
@@ -0,0 +1,111 @@
...
+### Deprecated: Binary Swift modules in dSYM bundles
```
davelester:
```suggestion
```

Removing in order to become bolded title in list item.

_posts/2026-08-25-module-tracking-in-debug-info.md
```diff
@@ -0,0 +1,111 @@
...
+As a consequence, `dsymutil` will no longer process binary Swift modules. This is a good thing, because binary Swift modules—which can only be parsed by the exact toolchain that produced them—were always at odds with dSYM bundles being a long-term archival format. Moreover, Swift modules often depend on Clang modules, and these Clang modules also were never included in dSYM bundles. By removing the binary Swift modules, dSYM bundles will get smaller.
```
davelester:
```suggestion
* __Binary Swift modules in dSYM bundles:__ As a consequence, `dsymutil` will no longer process binary Swift modules. This is a good thing, because binary Swift modules—which can only be parsed by the exact toolchain that produced them—were always at odds with dSYM bundles being a long-term archival format. Moreover, Swift modules often depend on Clang modules, and these Clang modules also were never included in dSYM bundles. By removing the binary Swift modules, dSYM bundles will get smaller.
```

_posts/2026-08-25-module-tracking-in-debug-info.md
```diff
@@ -0,0 +1,111 @@
...
+#### But don’t we need them for debugging?
```
davelester:
```suggestion
```

Removing in order to become a bolded title in nested list item.

_posts/2026-08-25-module-tracking-in-debug-info.md
```diff
@@ -0,0 +1,111 @@
...
+Since Swift 1.0, binary Swift modules were included in dSYM bundles because they were needed to resolve the types of local variables. However, starting with Swift 5.6, LLDB could perform this operation by reading the reflection metadata in the binary. The absence of binary Swift modules in dSYM bundles does not affect LLDB’s ability to inspect the contents of variables or dump object descriptions with `po`. Binary Swift modules are still needed to evaluate complex expressions like function calls or computed getters. Expression evaluation continues to work as long as LLDB finds all binary modules in their original (or [remapped](https://lldb.llvm.org/use/map.html#remap-source-file-pathnames-for-the-debug-session)) location. This is always the case when debugging a just-built binary on the same machine.
```
davelester:
```suggestion
   * __But don’t we need them for debugging?__ Since Swift 1.0, binary Swift modules were included in dSYM bundles because they were needed to resolve the types of local variables. However, starting with Swift 5.6, LLDB could perform this operation by reading the reflection metadata in the binary. The absence of binary Swift modules in dSYM bundles does not affect LLDB’s ability to inspect the contents of variables or dump object descriptions with `po`. Binary Swift modules are still needed to evaluate complex expressions like function calls or computed getters. Expression evaluation continues to work as long as LLDB finds all binary modules in their original (or [remapped](https://lldb.llvm.org/use/map.html#remap-source-file-pathnames-for-the-debug-session)) location. This is always the case when debugging a just-built binary on the same machine.
```

_posts/2026-08-25-module-tracking-in-debug-info.md
```diff
@@ -0,0 +1,111 @@
...
+If the absence of binary Swift modules in dSYM bundles creates an unforeseen problem with your workflow, please let us know, either on the [Swift LLDB forum](https://forums.swift.org/c/development/lldb/13) or by creating an issue on the [bug tracker](https://github.com/swiftlang/swift/issues)!
+### What about caching?
```
davelester:
```suggestion
```

Suggestion to remove this as a heading, and instead just have a concluding / clarifying sentence beneath the list items above. This may depend on the revision to the sentence itself.

_posts/2026-08-25-module-tracking-in-debug-info.md
davelester: I think a rewrite of his last sentence would help. Not as useful as a link about what CAS is, but how does this revision look?


```suggestion
The paths Swift stores for modules and debug info can now be content-addressed, identified by content rather than file location, so everything described here also works transparently with compilation caching.
```

_posts/2026-08-25-module-tracking-in-debug-info.md
davelester: I agree that more of an introduction would help. I've offered a suggestion for language that may introduce the reader to the topic: https://github.com/swiftlang/swift-org-website/pull/1501#discussion_r3953458066

_posts/2026-08-25-module-tracking-in-debug-info.md
adrian-prantl: Applied slightly modified.

_posts/2026-08-25-module-tracking-in-debug-info.md
adrian-prantl: yes, `-modulewrap` is correct.

_posts/2026-09-11-module-tracking-in-debug-info.md
adrian-prantl: the text below already calls out windows, I also added a reference to the paragraph above.

_posts/2026-08-25-module-tracking-in-debug-info.md
adrian-prantl: Reworded a little. Let me know if this works.

_posts/2026-09-11-module-tracking-in-debug-info.md
```diff
@@ -0,0 +1,125 @@
...
+---
+
+When your Swift program hits a breakpoint and stops so you can inspect it, the debugger's expression evaluator has to find the exact Swift module your code was built from. Until now, that lookup wasn't always precise. The upcoming Swift 6.4 release will include changes, begun in Swift 6.3, that address this by updating how the Swift compiler references explicitly-built Swift modules in debug info.
+
+The majority of developers will automatically benefit from **faster**, more **reliable debugging** and **smaller build products**, without any modifications to their SwiftPM or Xcode projects.
+
+For developers who **maintain their own build systems** using, for example, Bazel, Buck, or CMake, **some adjustments may be necessary** to take advantage of these changes.
+
+This article explains how Swift modules are used by the debugger, and how they are related to debugging. Next, it explains how Swift 6.3+ changes how modules are tracked in debug info to solve several problems with the previous representation. Finally, it shows how to adjust build systems to make use of the new representation, and eliminate some build steps that are no longer necessary.
```
davelester:
```suggestion
This article explains how the debugger uses Swift modules. Next, it describes how Swift 6.3 changes the way modules are tracked in debug info to solve several problems with the previous representation. Finally, it shows how to adjust build systems to take advantage of the new representation and eliminate some build steps that are no longer necessary.
```

davelester: Simplified language to make this clearer.

_posts/2026-09-11-module-tracking-in-debug-info.md
```diff
@@ -0,0 +1,125 @@
...
+
+* __Swift driver passes module path to compile jobs:__ Users of `swiftpm` or Xcode do not need to think about this, because the Swift driver also knows about the new `-debug-module-path` option and automatically passes the path to the object file's own Swift module to the compiler. However, users maintaining their own third-party build system to orchestrate Swift compilations with explicitly-built modules that are calling the Swift frontend directly and bypassing the Swift driver need to make sure to communicate the path to the top-level module to each object file compilation job.
+
+### What's deprecated
```
davelester, line 102:
```suggestion
### What's deprecated

_Beginning in Swift 6.4, you can safely make the following changes._

```

@adrian-prantl while the changes in "what's new" largely mention the version, it was unclear when these are deprecated, are they all deprecations in 6.4? If so, perhaps we can add a line like this suggestion.

_posts/2026-09-11-module-tracking-in-debug-info.md
adrian-prantl: Looks good.

_posts/2026-09-11-module-tracking-in-debug-info.md
adrian-prantl: Yes this captures the intention perfectly!

davelester (approved):

_posts/2026-09-11-module-tracking-in-debug-info.md
```diff
@@ -0,0 +1,128 @@
...
+
+## Summary
+
+With these changes for **explicitly-built modules**:
+- Binaries built with debug info on Windows and Linux, and dSYM bundles on Darwin will get dramatically smaller, since they no longer contain any binary Swift modules (6.4+)
+- Contextual module imports in LLDB become more reliable due to precise tracking instead of by-name lookups
+- Certain performance cliffs around module importing in LLDB are eliminated (such as SDK module dependencies in dSYMs triggering implicit imports)
+- **Developers maintaining their own build systems** can remove support for `-modulewrap` actions and remove `-add_ast_path` from the linker flags, but may need to pass `-debug-module-path` to the compiler if they are not letting the Swift driver handle the frontend options
+- Finally, many projects forgot to use `-add_ast_path` when linking static archives into their projects and had mysterious debugging problems inside the static archives as a result. This entire class of problems has been designed away.
```
davelester:
```suggestion
- Finally, static archives were easy to overlook: projects that didn't use `-add_ast_path` when linking them often had confusing debugging issues inside those archives as a result. This entire class of issues has been designed away.
```

_posts/2026-09-11-module-tracking-in-debug-info.md
```diff
@@ -0,0 +1,128 @@
...
+
+*tl;dr:* `-modulewrap` and `-add_ast_path` are replaced by `-debug-module-path`. Debug info gets smaller and more precise!
```
davelester:
```suggestion
*tl;dr:* `-modulewrap` and `-add_ast_path` are replaced by `-debug-module-path`. Debug info gets smaller and more precise.
```

_posts/2026-09-11-module-tracking-in-debug-info.md
```diff
@@ -0,0 +1,128 @@
...
+# Modulewrap and linker invocation on Linux
+swift-frontend -modulewrap Foo.swiftmodule -o Foo.swiftmodule.o
+lld Foo.o Foo.swiftmodule.o -o MyApplication
+```
+
+This can create scalability issues, especially for large applications:
+- Module files can get large and for an entire application you can often end up with a large portion of the SDK in the resulting binary. That can be quite problematic for the binary size.
+- As mentioned above, the chances of LLDB finding the right module in a Swift AST section or symbol table just by its base name diminish as the application gets more complex.
+- Binary Swift modules are version-locked to the precise compiler that created them. This is at odds with the intent of dSYM bundles, which are meant for long-term archival serialization of debug info.
+- When anything goes wrong and a matching explicit module cannot be found LLDB falls back to an implicit module import which may involve recompiling parts of the SDK from source. This can be very slow.
```
davelester:
```suggestion
- If a matching explicit module cannot be found, LLDB falls back to an implicit module import which may involve recompiling parts of the SDK from source. This can be very slow.
```

adrian-prantl: typo: "backto" otherwise this sounds fine

_posts/2026-09-11-module-tracking-in-debug-info.md
adrian-prantl: I like the tone of the suggestion.

_posts/2026-09-11-module-tracking-in-debug-info.md
adrian-prantl: lgtm

PR #1502, opened 2026-08-22
Fix performance percentage calculation in STTI chart

numist:
The arithmetic used to report relative performance percentage between Swift and C in the chart on the [Swift TrueType hinting interpreter blog post](https://www.swift.org/blog/migrating-truetype-hinting-to-swift/) contained a defect that made Swift sound slower than it actually was.


heckj (approved):

davelester:
@numist Thanks for sharing this! PR looks good, but I'm curious: does this change the "13% faster" metric that's mentioned twice in the post?

numist:
It does not!

davelester (approved):

